PR #2, opened 2019-01-10
[SL-1077] More reasonable types + minor cleanup

P0lip:


marbemac:
While you're at it, could you do simple search for some of these props to make sure they're still relevant? For example, this one can be removed: https://github.com/stoplightio/json-schema-viewer/blob/feat/SL-1077/convert-to-ts/src/JsonSchemaViewer.tsx#L18.

src/PropValidations.tsx
```diff
@@ -1,52 +1,46 @@
...
-        );
-      }
-    }
-
-    if (elems.length) {
-      return elems;
-    }
+    return (
+      <>
+        {Object.entries(validations).reduce(
```
wmhilton: I'd personally use `.map` + `.filter` just because it's less error prone. Actually you don't even need the filter because you can just return `null` for those elements and React won't render anything in those spots.

P0lip:
@marbemac sure thing. As you might have seen, I've already got rid of `forApiDocs` in the code, but simply forgot to remove it from the interface.
Will do the same for all irrelevant (unused) properties.

PR #3, opened 2019-01-13
[SL-1078] Setup theme

P0lip:
Apart from setting up the UI-Kit theme, I introduced a couple of unrelated changes.
These includes
- implemented slightly more generic error handling solution that relies on Error Boundaries rather than try...catch statements.
- optional schema parsing was pulled out of Schema Viewer and end project is expected to provide a parsed schema as of now.
- updated interfaces


P0lip:
@wmhilton 16.7 [was released without hooks](https://reactjs.org/blog/2018/12/19/react-v-16-7.html) :(

P0lip:
> Is the dereferencedSchema prop obsolete then? And follow on to that, all the related logic such as dereferenceSchema.ts?

Well, in fact we can remove it. We've already got more robust alternative in form of [json-ref-resolver](https://github.com/stoplightio/json-ref-resolver).

@marbemac addressed all concerns, feel free to re-review once you are some spare time.

PR #11, opened 2019-04-08
feat: jsv 2.0 [STD-95]

P0lip:
List of changes:
- removed all platform code related to schema rendering and wrote it once again (this a significant change, I didn't follow the old code at all and focused on spec)
- added dereferencedSchema back
- validations are displayed in a tooltip
- plenty of other changes

TODO:
- [x] `patternProperties` (see below)
- [ ] virtualization (separate PR)
- [x] row divider for `oneOf` / `anyOf`
- [x] `additionalItems` / `additionalProperties`
- [x] unique
- [x] enums
- [x] `$schema` and `$id` for root?
- [x] show more button
- [ ] tests (will update/write once all the open questions are answered)

Open questions:
- should we display patternProperties as regular object properties or validations? IMHO we should go with the former and display them as object properties with some annotation
- should we display validations in the tooltip? We went through that idea briefly last week, but no final decision has been made. I decided to implement the change, as I'm planning to add virtualization as a part of this PR.
- I am not quite sure how to deal with cases like:
```json
"type": ["array", "object", "number"],
"items": someItems
"properties": someProperties
```
The question is - should we display both items and properties or nothing? I tried to find the answer in the [spec](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.5.2.1) but to no avail. I don't even know if the shown syntax is in fact valid. At the moment we do display only properties and ignore items. Was it based on common sense or was there some actual spec?
- we are missing support for ['not'](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.5.6) - this is an oversight or it's just very uncommon and nobody really needs it?
- do we want to display actual schema for `additionalProperties` / `additionalItems`? It seems to be valid for both -  [5.3.1](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.3.1) [5.4.4](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.4.4)
- do we need `hideRoot` and `name`?

# Before
image

# After
image


wmhilton:
Nitpick - this line is too long:

image

wmhilton:
I liked the tooltips... for about 10 seconds. Then I realized...
- makes it much harder to read because I have to move my mouse constantly
- makes it impossible to print
- makes it hard to view on phone
- makes it hard to notice if a validation has changed between spec versions

so I'd ditch the tooltips

wmhilton:
Bug? `properties.ids` is not being rendered....

P0lip:
> Bug? properties.ids is not being rendered....

I left a note on that in the description. See the 3rd open question.

wmhilton:
I think the best solution is just to render `properties` if they are there (regardless of `type`), and to render `items` if they are there, and to possibly render both if they are both there.

src/JsonSchemaViewer.tsx
```diff
@@ -64,12 +63,10 @@ export class JsonSchemaViewer extends React.PureComponent<IJsonSchemaViewer, IJs
           emptyText={emptyText}
           defaultExpandedDepth={defaultExpandedDepth}
           expanded={expanded}
-          hideInheritedFrom={hideInheritedFrom}
           hideRoot={hideRoot}
```
marbemac: We can probably ditch this property - it's a leftover from an old use case.

src/SchemaView.tsx
```diff
@@ -1,48 +1,57 @@
...
     emptyText,
     expanded = false,
-    hideInheritedFrom = false,
     hideRoot,
-    limitPropertyCount = 0,
+    limitPropertyCount,
     schema,
-    schemas = {},
+    dereferencedSchema,
+    originalSchema,
```
marbemac: Where is this used? Don't see it in `ISchemaView`.

marbemac: Where is this used? Don't see it in `ISchemaView`

src/hooks/useProperties.ts
```diff
@@ -0,0 +1,30 @@
...
+  options: IFilterOptions
+) => {
+  return useMemo(
+    () => {
+      let { limitPropertyCount } = options;
+      const nodes: SchemaTreeNode[] = [];
+      let isOverflow = false;
+
+      for (const node of getProperties(schema, dereferencedSchema, options)) {
+        if (limitPropertyCount !== undefined && limitPropertyCount-- < 0) {
```
marbemac: Mutating `limitPropertyCount` in the `if` statement smells a bit. Suggest putting on it's own line.

P0lip:
> any particular reason to split components between common and components folders?

Nope, it's a leftover. I was actually planning to move them under `components`.

src/SchemaView.tsx
P0lip: I believe it's a leftover. I need that for #12. Will remove.

PR #14, opened 2019-04-18
feat: integrate tree-list component [STU-123]

P0lip:
```
 src/__fixtures__/stress-schema.json                         | 15432 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
Don't be scared by diff, it's mostly new fixtures.


marbemac:
@P0lip is the mask stuff opt in? e.g. basically hidden right now, until we pass a prop or something in to show it when we're ready, in the right context (when we tackle the rest of masking feature in second half of may)? If not, let's make it opt in so that it doesn't disrupt anything right now.

P0lip:
@marbemac 
It's somewhat opt-in.
In order to have `edit mask` button displayed, you need to provide correct dereferenced schema.
If you don't pass it, the button won't be displayed and no masked schema will be ever rendered in a dialog unless you pass some extra properties.

PR #16, opened 2019-05-07
fix: handler combiners with type

P0lip:


src/utils/renderSchema.ts
```diff
@@ -80,6 +80,10 @@ export const renderSchema: Walker = function*(schema, dereferencedSchema, level
       if (node.properties !== undefined) {
         const isConditionalCombiner = node.combiner === 'anyOf' || node.combiner === 'oneOf';
```
casserni: add allOf here?

P0lip: Sounds good! Will make the change in a bit.

PR #17, opened 2019-05-08
fix: array of objects with properties

P0lip:


src/utils/walk.ts
```diff
@@ -33,20 +33,28 @@ function assignNodeSpecificFields(base: IBaseNode, node: JSONSchema4) {
   }
 }
 
+function getType(node: JSONSchema4) {
+  if (Array.isArray(node.type)) {
+    return node.type.length === 1 ? node.type[0] : node.type;
```
lottamus: We return a string if there's only one, but an array if theres none or more than one? Is that correct? Just making sure

lottamus:
Do we have this fix on the v2 branch as well?

src/utils/walk.ts
P0lip: > We return a string if there's only one, but an array if theres none or more than one? Is that correct? Just making sure

Well, I'm not sure how to handle such a scenario. For instance, there is no way we can display both `array` of items and `object` with properties.
Our UI is just too simple.

We would need to have something more or less similar to http://jlblcc.github.io/json-schema-viewer/ to handle such a case.

@marbemac what your thought on that?

P0lip: Also, we should be handling `not`, which we don't at the moment.
Any suggestions on the UI?

lottamus: > there is no way we can display both array of items and object with properties

Is this what you mean?

image


Or something like `oneOf`?

image

marbemac: What is `not`?

We don't have to handle array and object in same property. We can suggest people use `oneOf` if they need that, although its kind of crazy and definitely an edge case.

P0lip: https://json-schema.org/understanding-json-schema/reference/combining.html#not

P0lip: @lottamus yeah, pretty much that
```json
{
  "type": ["array", "object"],
  "properties": {
    "bar": { 
      "type": "string" 
    }
  },
  "items": [
     {
       "type": "number"
     },
     {
       "type": "object",
       "properties": {
          "foo": { 
            "type": "number" 
          }
        }
     }
  ]
}

marbemac: Got it, we can skip `not` for now - I haven't seen it used in the wild by our customers yet, and would need to support it in JSE as well. A good addition for later, but not high priority right now.

PR #18, opened 2019-05-15
 feat: primary type

P0lip:


src/utils/walk.ts
```diff
@@ -84,7 +82,20 @@ function processNode(node: JSONSchema4): SchemaNode | void {
     } as IRefNode;
   }
 
-  if (combiner !== undefined) {
+  if (combiner === undefined) {
+    const base: IBaseNode = {
+      id: assignId(node),
+      primaryType: getPrimaryType(node),
+      type: node.type || inferType(node),
```
marbemac: Why have two properties? When will a node ever have a defined type AND a primary type?

marbemac: Any reason we can't just update `inferType` and be done with it?

P0lip: `primaryType` - this props drives the rendering logic
`type` -  displayed in UI

src/utils/isSchemaViewerEmpty.ts
```diff
@@ -1,6 +1,5 @@
 import { ISchema } from '@stoplight/types';
-import get = require('lodash/get');
-import isEmpty = require('lodash/isEmpty');
+import { get, isEmpty } from 'lodash';
```
casserni: _get and _isEmpty

src/utils/walk.ts
```diff
@@ -33,41 +33,39 @@ function assignNodeSpecificFields(base: IBaseNode, node: JSONSchema4) {
...
 
-  if (node.type !== undefined && combiner === undefined) {
-    const base: IBaseNode = {
-      id: assignId(node),
-      type: getType(node),
-      validations: getValidations(node),
-      annotations: getAnnotations(node),
-      enum: node.enum,
-    };
+function getPrimaryType(node: JSONSchema4) {
```
casserni: what if type is `["object", "array"]`

src/utils/walk.ts
casserni: Does primary type need to be store on the node if not used in the ui? Can we just call `getPrimaryType` in renderSchema when its needed

src/utils/walk.ts
casserni: Really the user should be using an oneOf like below, but if having both in type is valid we definitely need to support it

```
{
    "oneOf": [
        {"$ref": "#/definitions/singleObject"}, // plain object
        {
            "type": "array", // array of plain objects
            "items": {"$ref": "#/definitions/singleObject"}
        }
    ],
    "definitions": {
        "singleObject": {...}
    }
}```

casserni: This also means that if type is `["object", "array"]` we need to infer that its a combiner type oneOf so it renders properly in the ui

P0lip: Yeah, shared the same concern here https://github.com/stoplightio/json-schema-viewer/pull/17#discussion_r283318479.
I'm not sure if displaying `oneOf` is safe though. To me, it sounds like a nice idea, but we'll need to make it reliable (transform the data correctly).
 @marbemac, what's your take on this?

src/utils/walk.ts
P0lip: Fair enough!

src/utils/walk.ts
marbemac: I think we don't worry about this use case until a customer complains about it.

PR #22, opened 2019-05-17
Feat/dark

casserni:


src/components/JsonSchemaViewer.tsx
```diff
@@ -70,7 +70,16 @@ export class JsonSchemaViewerComponent extends React.PureComponent<IJsonSchemaVi
       return <div>{emptyText}</div>;
     }
 
-    return <SchemaTree expanded={expanded} name={name} schema={schema} treeStore={this.treeStore} {...props} />;
+    return (
+      <SchemaTree
+        className="JsonSchemaViewer"
```
lottamus: This will break any existing classnames that are passed in. Can you wrap this in a cn?

```suggestion
        className={cn("JsonSchemaViewer", className)}
```

PR #24, opened 2019-05-21
feat: ref details popover [STU-185]

P0lip:
See the STU-180 story for more info.

Light theme:
Screenshot_20190521_222134
Dark theme:
Screenshot_20190521_221951

Open questions:
- is tooltip-based approach a right call?
- should we make URL a bit more flexible and allow it to be, say, a callback? it might make routing easier in the future.
- how would the Studio integration look like?


lottamus:
Could just do a simple inline "(go to ref)" like we had in Next? 

Clicking "(go to ref)" could emit `JsonSchemaEditorStoreEvents.GoToRef` or something and let the consumer handle it however. 

Studio can listen to that event, look up a Model node that matches the given URI, and set it as the active node?

image

marbemac:
I agree with @lottamus - what do you think @P0lip?

P0lip:
@marbemac 
Already implemented in https://github.com/stoplightio/json-schema-editor/pull/24
Shall we apply the same change to JSV?

lottamus:
> Shall we apply the same change to JSV?

Yeah I'd say so. But if we're showing a `$ref` in JSV it's likely because we couldn't resolve it so its unlikely that the `$ref` even exists

PR #36, opened 2019-07-10
fix incorrect rendering of array of objects with no type [STU-362]

mallachari:
Related to [STU-362](https://stoplightio.atlassian.net/browse/STU-362)

Changes:
- fix issue causing not rendering schema if object node does not have a type specified

Context:
Issue appeared only for arrays that contained objects without specified type. This way node subtype property was undefined and the process of node children rendering stopped.  

Before:
Screenshot from 2019-07-12 12-57-08

After:
Screenshot from 2019-07-12 12-55-59

### DIFF
--- a/src/utils/renderSchema.ts
+++ b/src/utils/renderSchema.ts
@@ -77,7 +77,12 @@ export const renderSchema: Walker = function*(schema, level = 0, meta = { path: 
         ...meta,
         ...(parsedSchema.items !== undefined &&
           !Array.isArray(parsedSchema.items) && {
-            subtype: '$ref' in parsedSchema.items ? `$ref( ${parsedSchema.items.$ref} )` : parsedSchema.items.type,
+            subtype:
+              '$ref' in parsedSchema.items
+                ? `$ref( ${parsedSchema.items.$ref} )`
+                : parsedSchema.items.type ||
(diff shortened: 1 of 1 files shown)


src/utils/renderSchema.ts
```diff
@@ -77,7 +77,12 @@ export const renderSchema: Walker = function*(schema, level = 0, meta = { path:
         ...meta,
         ...(parsedSchema.items !== undefined &&
           !Array.isArray(parsedSchema.items) && {
-            subtype: '$ref' in parsedSchema.items ? `$ref( ${parsedSchema.items.$ref} )` : parsedSchema.items.type,
+            subtype:
+              '$ref' in parsedSchema.items
+                ? `$ref( ${parsedSchema.items.$ref} )`
+                : parsedSchema.items.type ||
+                  (parsedSchema.items.properties && 'object') ||
```
P0lip: `(parsedSchema.items.properties && 'object')`
The RHS will always be truthy.

P0lip:
This PR will need to be revamped a bit, as we have `inferType` which should be used here.
I'll refactor this tomorrow.

Suggest adding Chris or Nick to the PRs, they have a fairly decent experience with JSV, so they might spot such nuances.

cc @mallachari @chris-miaskowski

PR #37, opened 2019-07-10
feat: introduce rowRendererRight

chris-miaskowski:
## What change is introduced

- added `rowRendererRight` agnostic of the content (will enable Studio to render checkboxes). This component renders arbitrary elements on the right hand side of the row.
- added `schemaControlsRenderer` agnostic of the content (will enable Studio to render 'apply mask' and 'clear mask' buttons). This renders any content right under the JSON Schema
- fix for combiner dividers. previously it was relative to the immediate parent which cause an issue if `rowRendererRight` had any margin or modified height in any way. it resulted in that parent's calculated height to be altered and therefore the relative positionig of divider become broken. We had to use `calc` but it's supported by all major browsers.

See a storybook with example checkbox added.

image


src/components/SchemaTree.tsx
```diff
@@ -50,9 +42,22 @@ export const SchemaTree = observer<ISchemaTree>(props => {
         striped
         maxRows={maxRows !== undefined ? maxRows + 0.5 : maxRows}
         store={treeStore}
-        rowRenderer={rowRenderer}
+        rowRenderer={(node, rowOptions) => {
```
chris-miaskowski: We had to remove the optimization because `rowRendererRight` was not rerendered otherwise.

src/utils/renderSchema.ts
```diff
@@ -67,7 +67,7 @@ export const renderSchema: Walker = function*(schema, level = 0, meta = { path:
 
   const { path } = meta;
 
-  for (const node of walk(parsedSchema)) {
+  for (const node of walk(resolvedSchema)) {
```
mallachari: What is the reason for that change?

src/__stories__/JsonSchemaViewer.tsx
```diff
@@ -6,6 +6,7 @@ import { boolean, number, object, select, text, withKnobs } from '@storybook/add
 import { storiesOf } from '@storybook/react';
 import { JsonSchemaViewer } from '../components';
 
+import { Checkbox } from '@blueprintjs/core';
```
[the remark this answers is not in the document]
chris-miaskowski: will be changed in following commit

src/utils/renderSchema.ts
chris-miaskowski: good point, development artifact, will revert

src/components/SchemaRow.tsx
```diff
@@ -8,21 +8,27 @@ import get = require('lodash/get');
...
 const ICON_SIZE = 12;
 const ICON_DIMENSION = 20;
-const ROW_OFFSET = 7;
 
-export const SchemaRow: React.FunctionComponent<ISchemaRow> = ({ node, rowOptions, onGoToRef }) => {
+export const SchemaRow: React.FunctionComponent<ISchemaRow> = ({
+  node,
+  rowOptions,
+  onGoToRef,
+  rowRendererRight,
```
P0lip: This is kind of a hack. Ideally, the entire row should be customizable.

src/components/SchemaTree.tsx
P0lip: It shouldn't be breaking anything. You just need to specify the dependency for useCallback

src/components/SchemaRow.tsx
chris-miaskowski: That's the solution suggested by @marbemac. I personally don't think it's much of a hack, but a hook :) Similarly to how you'd get plugins a space to render buttons. We can make this more versatile of course in the future, but making the row ultra generic wasn't the plan.

Side note: I don't like the idea of making the whole row customizable. I mean I see the merit of that but I don't think that we need so much flexibility.

src/components/SchemaTree.tsx
chris-miaskowski: I agree that it shouldn't but it does. The problem is that: the underlying `rowRendererRight` doesn't re-render (which is, by the way, passed as a React.callback as well). I'm happy to jump on a call to explain what's broken (long story short: checkboxes don't rerender), but I couldn't find any workaround (and passing checkbox's state is not an option).

src/components/SchemaRow.tsx
P0lip: He suggested 2 solutions - a simple one and also a bit more difficult one but also the better one.
Obviously it's by far better in comparison to the previous solution but having a possibility to customise the entire row would be make the solution more flexible. 
This is just some sort of workaround. I'm fine with that, don't worry, but would really wish it was more flexible. Tree list also exposed such a simple renderer prop initially, but we ended up with the entire row being customizable. The same will likely to apply here at some point, but it's good as is now.

src/components/SchemaTree.tsx
P0lip: Just leave a todo comment here, I'll revisit it later. Nothing blocking this PR though.

P0lip:
I reviewed the code and it's alright. If the component looks okay, we can merge it. I'm not fully happy with that renderer prop, yet I won't reject PR because of this.

src/components/SchemaTree.tsx
lag-of-death: We could use `useCallback` having it depend on https://github.com/stoplightio/studio-internal/blob/e65356b86f4162c5ba51ebd0c0b4e219c2ca4e2b/src/components/MaskEditor/index.tsx#L109 but this would mean passing `maskProps` somehow to JSV

lag-of-death: @P0lip, I've left the TODO. UI is looking fine to me. cc @mallachari

src/components/SchemaRow.tsx
chris-miaskowski: Yup, if we need more flexibility I'm all for it and can even do the refactoring myself if help's needed. I think what @marbemac suggested is to go with the simpler suggestion now and I put my bet on it since I think it makes sense for now. Anyway, looks like we're on alignment.

chris-miaskowski:
> I reviewed the code and it's alright. If the component looks okay, we can merge it. I'm not fully happy with that renderer prop, yet I won't reject PR because of this.

Cool! Thank you, can you please hit approve in such case? Would get one dependency ticked off.

src/components/SchemaRow.tsx
P0lip: Yeah, it's totally okay for now :)

marbemac: I know it's merged, but a quick follow on here (for later PR consideration etc). Apologies if this was already the idea, but:

- Could add the `rowRenderer(props, opts: { DefaultRow: ReactElement })` prop.
- Extract the current elements in the row into a standalone component, expose that as that  `DefaultRow` property passed to rowRenderer opts.
- Up to user of rowRenderer to render `<DefaultRow {...props} />` somewhere if the want it, and then can tweak the props to it (or just pass them through as is), and put whatever they want around it.

Shouldn't be too difficult. For future consideration.

P0lip: Yessir! This is *exactly* what I wanted to see being implemented here and what I suggested in the previous PR. 
I'll do it once I have a bit of spare time

PR #40, opened 2019-07-13
feat: introduce rowRenderer 

P0lip:
List of changes:
- working optional allOf merging
- rowRenderer instead of `rowRendererRight` (please refer to https://github.com/stoplightio/json-schema-viewer/pull/37#discussion_r302609270 for more details)
- cleanup

Needed to unyalc JSV in Studio.

PR #41, opened 2019-07-15
refactor: pass nodes to rowRendererRight

lag-of-death:
This gives more control to `rowRendererRight`. Thanks to this change, it is now possible to compare the current node with the rest of the nodes that JSV has access to.

### DIFF
--- a/src/components/SchemaRow.tsx
+++ b/src/components/SchemaRow.tsx
@@ -8,15 +8,16 @@ import get = require('lodash/get');
 import map = require('lodash/map');
 import size = require('lodash/size');
 
-import { GoToRefHandler, IExtendableRenderers, SchemaNodeWithMeta, SchemaTreeListNode } from '../types';
+import { GoToRefHandler, SchemaNodeWithMeta, SchemaTreeListNode } from '../types';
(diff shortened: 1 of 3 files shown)


P0lip:
Keep in mind `rowRendererRight` is about to be nuked. https://github.com/stoplightio/json-schema-viewer/pull/40 - going to finish this today and hopefully it gets merged very soon.
I suggest closing this PR and staying with yalc'ed version in Studio to avoid triggering yet another release.

chris-miaskowski:
@P0lip @lag-of-death 
Since this is close should we expect that the changes @P0lip is working on will include changes @lag-of-death introduced here?

lag-of-death:
@chris-miaskowski, I think that's more for @P0lip to say actually since he's working on https://github.com/stoplightio/json-schema-viewer/pull/40. Thanks for keeping me in the loop

P0lip:
> Since this is close should we expect that the changes @P0lip is working on will include changes @lag-of-death introduced here?

Yeah, I can expose nodes somehow, if needed.

lag-of-death:
@P0lip, yeah, it is needed :+1:

Issue #42, opened 2019-08-03
Changing observed observable values outside actions is not allowed


lottamus:
Platform Studio is complaining that we're changing an observed observable's value outside an action.

![image](https://user-images.githubusercontent.com/7423098/62413420-73ae2e80-b5d4-11e9-8883-808a25d24708.png)

![image](https://user-images.githubusercontent.com/7423098/62413428-86c0fe80-b5d4-11e9-9ac4-d4cf98f5c3ab.png)

P0lip:
Should be fixed in https://github.com/stoplightio/json-schema-viewer/issues/42

PR #43, opened 2019-08-04
feat: perform full processing in a worker

P0lip:
Fixes https://github.com/stoplightio/json-schema-viewer/issues/42
Closes https://github.com/stoplightio/json-schema-viewer/issues/45


.storybook/webpack.config.js
```diff
@@ -1,7 +1,11 @@
 const defaultConfig = require('@stoplight/storybook-config/webpack.config');
+const path = require('path');
```
wmhilton:
```suggestion
const path = require('@stoplight/path');
```
?

P0lip: I don't think it's installed, yet can do.

babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  'plugins': ['@babel/plugin-transform-modules-commonjs'],
```
wmhilton: can you write a paragraph explaining why we need to add Babel to the ever-growing-list-of-compilers we need to support? Looks like it's only used by Jest?

wmhilton: hang on I just realized this isn't `studio-internal` 😆 um... I'm still curious though

P0lip: It's because of of `lodash-es`. I suspect I should be able to instrument TS to transpile those files, though, so will give it a try.

marbemac: I see we replaced `lodash` - what is `lodash-es` doing for us?

P0lip: Rollup does not really like CJS modules, so this is the primary reason.
Lodash-es is mostly used in a worker code, so it does not really change much now.

P0lip:
> a very subtle off-by-one thingy.... the pre-rendered one didn't render the "half-row" at the bottom

That's a feature? We want to have that cut in order to indicate the content can be scrolled

PR #46, opened 2019-08-28
fix: array subtype

mallachari:
Fixes [#71](https://github.com/stoplightio/studio/issues/71)

Changes:
- checks node type before inferring 
- adds tests

Screenshot from 2019-08-28 16-29-00

### DIFF
--- a/src/components/__tests__/Property.spec.tsx
+++ b/src/components/__tests__/Property.spec.tsx
@@ -1,0 +1,27 @@
+import { shallow } from 'enzyme';
+import 'jest-enzyme';
+import * as React from 'react';
+import { SchemaNodeWithMeta } from '../../types';
+import { Property, Types } from '../shared';
+
+describe('Property component', () => {
+  it('should render Types with propper type and subtype', () => {
+    const node: SchemaNodeWithMeta = {
+      id: '1',
+      type: 'array',
+      items: {
+        type: 'string',
+      },
+      annotations: {
+        examples: {},
+      },
+      validations: {},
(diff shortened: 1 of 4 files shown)


src/utils/getNodeType.ts
```diff
@@ -0,0 +1,6 @@
+import { JSONSchema4, JSONSchema4TypeName } from 'json-schema';
+import { inferType } from './inferType';
+
+export function getNodeType(node: JSONSchema4): JSONSchema4TypeName | JSONSchema4TypeName[] | undefined {
```
P0lip: Let's just extend `inferType` function.

src/components/__tests__/Property.spec.tsx
```diff
@@ -0,0 +1,27 @@
+import { shallow } from 'enzyme';
+import 'jest-enzyme';
+import * as React from 'react';
+import { SchemaNodeWithMeta } from '../../types';
+import { Property, Types } from '../shared';
+
+describe('Property component', () => {
+  it('should render Types with propper type and subtype', () => {
```
P0lip: :+1:

src/utils/getNodeType.ts
mallachari: That was actually my first intention but thought it might not be precise enough. 
I'll move it there. But if so shouldn't `type` be checked before `properties` and `items`?

P0lip: Yeah, it could be moved to the top.

Issue #49, opened 2019-09-05
Uncaught error causes endless spinner


lottamus:
**Describe the bug**

As expected, adding an allOf with different types logs the following error to the console:

> Uncaught Error: Could not resolve values for path:"type". They are probably incompatible. Values: "string" "number"

The UI displays an endless spinner with no information about the error.

**To Reproduce**

1. Go to https://stoplightio.github.io/json-schema-viewer/?path=/story/jsonschemaviewer--custom-schema
2. Add the following schema:

```json
{
   "title": "All Of",
   "allOf": [
      {
         "type": "string"
      },
      {
         "type": "number"
      }
   ]
}
```

**Expected behavior**

An error boundary is shown displaying the error.

**Screenshots**

![image](https://user-images.githubusercontent.com/7423098/64302620-cb301900-cf49-11e9-9f21-5ff43f54358e.png)

PR #51, opened 2019-09-05
fix(rendering): uncaught exception

P0lip:
Fixes https://github.com/stoplightio/json-schema-viewer/issues/49

### DIFF
--- a/src/workers/schema.ts
+++ b/src/workers/schema.ts
@@ -1,17 +1,26 @@
 import { mergeAllOf } from '../utils/mergeAllOf';
 import { renderSchema } from '../utils/renderSchema';
-import { ComputeSchemaMessage } from './messages';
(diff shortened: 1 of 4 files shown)


wmhilton:
oh i guess i should take a look at this.

src/workers/schema.ts
```diff
@@ -1,17 +1,26 @@
...
+      instanceId,
+      error: null,
+      nodes: Array.from(renderSchema(shouldMergeAllOf ? mergeAllOf(schema) : schema)),
+    } as RenderedSchemaSuccessMessageData);
+  } catch (ex) {
+    self.postMessage({
+      instanceId,
+      error: ex.message,
+      nodes: null,
+    } as RenderedSchemaErrorMessageData);
```
wmhilton: 🤢 Think it is time to pull out `MagicPortal` back into its own repo so we can have some of that goodness here?

P0lip: @wmhilton yeah, that would be great

Issue #58, opened 2019-10-15
JSON schema viewer should not mutate


marbemac:
**Describe the bug**

See title, it's causing issues elsewhere. There is at least one spot:

https://github.com/stoplightio/json-schema-viewer/blob/master/src/utils/renderSchema.ts#L107

PR #59, opened 2019-10-15
fix: do not mutate combiner child

P0lip:
Fixes https://github.com/stoplightio/json-schema-viewer/issues/58
A note on snapshot change - it shouldn't have any impact, since the actual node that gets rendered in tree-list has a proper type assigned - see [line 146](https://github.com/stoplightio/json-schema-viewer/blob/b186ec373f0b3f00184f4171ce3412b4b5987cb2/src/utils/__tests__/__snapshots__/renderSchema.spec.ts.snap#L146) and [line 197](https://github.com/stoplightio/json-schema-viewer/blob/b186ec373f0b3f00184f4171ce3412b4b5987cb2/src/utils/__tests__/__snapshots__/renderSchema.spec.ts.snap#L197).
It's just that the other nodes point at original object.

You can also verify it manually in storybook by pasting the schema file and comparing the view. :P

This got me realize we actually seem to squeeze too much into metadata. It seems like we could make it much leaner.

### DIFF
--- a/src/utils/renderSchema.ts
+++ b/src/utils/renderSchema.ts
@@ -104,10 +104,13 @@ export const renderSchema: Walker = function*(schema, level = 0, meta = { path: 
       if (node.properties !== undefined) {
         for (const [i, property] of node.properties.entries()) {
           if ('type' in node) {
-            property.type = property.type || node.type;
+            node.properties[i] = {
+              ...property,
+              type: property.type || node.type,
+            };
           }
 
-          yield* renderSchema(property, level + 1, {
+          yield* renderSchema(node.properties[i], level + 1, {
             ...(i !== 0 && { divider: DIVIDERS[node.combiner] }),
             path: [...path, node.combiner, i],
           });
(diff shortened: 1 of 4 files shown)


src/utils/renderSchema.ts
```diff
@@ -104,10 +104,13 @@ export const renderSchema: Walker = function*(schema, level = 0, meta = { path:
       if (node.properties !== undefined) {
         for (const [i, property] of node.properties.entries()) {
           if ('type' in node) {
-            property.type = property.type || node.type;
+            node.properties[i] = {
```
marbemac: So is this needed in general? Feels like it might not be.

P0lip: Yeah, it's needed. If you revert it, a few tests covering mutation will fail.

marbemac: Are you sure? For me only two tests fail, both snapshots regarding the "type": "object" thing. Not saying that means everything is good, but just that the tests don't seem to indicate any big issues?

Screen Shot 2019-10-15 at 6 58 00 PM

P0lip: >  For me only two tests fail,

Yup, these are the only ones, since they cover the combiner children case.

> Not saying that means everything is good, but just that the tests don't seem to indicate any big issues?

That's right, seems like this was the only spot.

PR #63, opened 2019-12-27
feat: export utils functions

mallachari:
Changes:
- exports `renderSchema` and `mergeAllOf`.


P0lip:
Do keep in mind that even though these functions are exposed now, we should not be using them in the end code. The implementation is internal and tied to tree list, therefore should not be really used elsewhere. I'm planning to make a significant change very soon, so any code that uses them would, most likely, need to be updated.

PR #64, opened 2020-01-29
feat: render deprecated next to general validations

P0lip:
Addresses https://github.com/stoplightio/platform-internal/issues/736
Before:
image

After:
image

### DIFF
--- a/src/components/shared/Validations.tsx
+++ b/src/components/shared/Validations.tsx
@@ -5,10 +5,13 @@ import * as React from 'react';
 
 export interface IValidations {
   required: boolean;
-  validations: Dictionary<unknown> | {};
+  validations: (Dictionary<unknown> | {}) & { deprecated?: boolean };
 }
 
-export const Validations: React.FunctionComponent<IValidations> = ({ required, validations }) => {
+export const Validations: React.FunctionComponent<IValidations> = ({
+  required,
+  validations: { deprecated, ...validations },
+}) => {
   const validationCount = Object.keys(validations).length;
 
   const requiredElem = (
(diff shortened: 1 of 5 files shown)


src/components/shared/Validations.tsx
```diff
@@ -18,49 +21,54 @@ export const Validations: React.FunctionComponent<IValidations> = ({ required, v
...
+                  elem = validation.map((v, i) => (
+                    <div key={i} className="mt-1 mr-1 flex items-center">
+                      <div className="px-1 bg-gray-2 dark:bg-gray-8 font-bold text-sm rounded">{String(v)}</div>
+                      {i < validation.length - 1 ? <div>,</div> : null}
+                    </div>
+                  ));
+                } else if (typeof validation === 'object') {
+                  elem = (
+                    <div className="m-1 px-1 bg-gray-2 dark:bg-gray-8 font-bold text-sm rounded" key={index}>
+                      {'{...}'}
```
lottamus: Why don't we stringify the validation here too so we can display the properties? Similar to what were doing on L53

src/components/shared/Validations.tsx
```diff
@@ -18,49 +21,54 @@ export const Validations: React.FunctionComponent<IValidations> = ({ required, v
...
-    <Popover
-      boundary="window"
-      interactionKind="hover"
-      content={
-        <div className="p-5" style={{ maxHeight: 500, maxWidth: 400 }}>
-          {Object.keys(validations).map((key, index) => {
-            const validation = validations[key];
+  return (
+    <>
+      {deprecated ? <span className="ml-2 text-darken-7 dark:text-lighten-6">deprecated</span> : null}
```
lottamus:
```suggestion
      {deprecated ? <span className="ml-2 text-orange-7 dark:text-orange-6">deprecated</span> : null}
```

src/components/shared/Validations.tsx
```diff
@@ -18,49 +21,54 @@ export const Validations: React.FunctionComponent<IValidations> = ({ required, v
...
+                } else if (typeof validation === 'object') {
+                  elem = (
+                    <div className="m-1 px-1 bg-gray-2 dark:bg-gray-8 font-bold text-sm rounded" key={index}>
+                      {'{...}'}
+                    </div>
+                  );
+                } else {
+                  elem = (
+                    <div className="m-1 px-1 bg-gray-2 dark:bg-gray-8 font-bold text-sm rounded" key={index}>
+                      {JSON.stringify(validation)}
```
lottamus: Should we safe stringify here?

src/components/shared/Validations.tsx
P0lip: I think the reason we did it in the first place is that the object might be large and therefore the text produced wouldn't be readable.

src/components/shared/Validations.tsx
P0lip: I don't know. I'd say no, since you won't be able to recognize strings and numbers, i.e. "1" and 1.

Example:
- JSON.stringify
  - string 2
    image
  - number 2
image

- safe stringify
  - string 2
image
  - number 2
image

PR #65, opened 2020-02-04
fix: handle nullish items

P0lip:
Fixes https://github.com/stoplightio/studio-internal/issues/1608

### DIFF
--- a/src/components/shared/Property.tsx
+++ b/src/components/shared/Property.tsx
@@ -1,6 +1,7 @@
-import { isObjectLike as _isObjectLike, size as _size } from 'lodash-es';
+import { size as _size } from 'lodash-es';
 import * as React from 'react';
 import { GoToRefHandler, IArrayNode, IObjectNode, SchemaKind, SchemaNodeWithMeta } from '../../types';
+import { isArrayNodeWithItems } from '../../utils/guards';
 import { inferType } from '../../utils/inferType';
 import { isCombinerNode, isRefNode } from '../../utils/nodes';
 import { Types } from './Types';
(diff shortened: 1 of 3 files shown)


src/components/shared/Property.tsx
```diff
@@ -13,7 +13,7 @@ export interface IProperty {
 export const Property: React.FunctionComponent<IProperty> = ({ node, onGoToRef }) => {
   const type = isRefNode(node) ? '$ref' : isCombinerNode(node) ? node.combiner : node.type;
   const subtype =
-    type === SchemaKind.Array && (node as IArrayNode).items !== undefined
+    type === SchemaKind.Array && _isObjectLike((node as IArrayNode).items)
```
marbemac: Should we start creating some type guards, like `isArrayNode`, and using those?

Any other spots where we might end up w unexpected null. `properties`, `anyOf`, `allOf`, `oneOf`, etc?

P0lip: > Should we start creating some type guards, like isArrayNode, and using those?

Yeah, sounds indeed a look a good idea. Will do.

> Any other spots where we might end up w unexpected null. properties, anyOf, allOf, oneOf, etc?

Yes there might be other spots, _but_ we take nulls into account in a few places, but just not all.

P0lip:
Done
In general typings need some love in json-schema-viewer.
It became quite a bit of a hassle once we started addressing all the weirdness and quirkiness.
`chore/typings` branched my work off here, gonna polish it one day, mostly to utilize tagged union literals and being more careful about malformed data.
I believe we need to take a similar approach as in case of http-spec and assume invalid data is passed rather than the data meets our expectations.

marbemac:
> I believe we need to take a similar approach as in case of http-spec and assume invalid data is passed rather than the data meets our expectations.

Or we add a transformer to `http-spec` package to "clean/normalize" json schema to make it valid, so that our entire ecosystem doesn't have to worry about these issues.

P0lip:
> Or we add a transformer to http-spec package to "clean/normalize" json schema to make it valid, so that our entire ecosystem doesn't have to worry about these issues.

That would be some solution as well, but wondering how it'd impact the perf of large schemas processing. If the hit isn't significant, I'd be all in for bringing it to JSV, as right now we kind of need to do the same stuff twice.

PR #66, opened 2020-02-05
fix: don’t kill entire JSV for one null value

marbemac:
Open to suggestions on better alternative. Current behavior is no good - if there is a `null` value the entire component doesn't work.

@P0lip thoughts?


src/utils/renderSchema.ts
```diff
@@ -47,9 +47,9 @@ const getPatternProperties: Walker = function*(schema, level = 0, meta) {
 
 export const renderSchema: Walker = function*(schema, level = 0, meta = { path: [] }, options = {}) {
   if (typeof schema !== 'object' || schema === null) {
-    throw new TypeError(
-      `Expected schema to be an "object" but received ${schema === null ? '"null"' : `a "${typeof schema}"`}`,
-    );
+    const error = `Expected schema to be an "object" but received ${schema === null ? '"null"' : `a "${typeof schema}"`} at path ${JSON.stringify(meta.path)}`;
+    console.warn(error);
```
P0lip: I guess we don't need to log it at all?
Doesn't give us much.
```suggestion
```

PR #69, opened 2020-02-14
feat: json-schema-viewer v3

P0lip:
~Tests need to be updated, but apart from that everything besides merge-all-of should be working.~

List of changes:
- local and external $refs can be expanded
image
- new tree-list
- improved performance


marbemac:
Ok so this is more of a tree-list thing, but since we don't have a PR for tree list yet, putting it here.

Curious if below makes sense to you, and what you think? I admittedly do not know as much about all the considerations for tree list, such as the `walk` generator, etc.

### Potential Data Loading Strategy

Tree data is built up by calling the user provided `computeRows` function.

`computeRows` should only return a single level of rows, not nested / deep.

```ts
type TreeNode<M = unknown> = {
  id: string;
  name: string;
  parent: TreeParentNode | null;
  type?: string;
  metadata: M;
}

type TreeParentNode<M = unknown, C extends TreeNode = TreeNode<M>> = TreeNode<M> & {
  children: C[];
}

export type TreeRow<M = unknown> = {
  id: string;
  name: string;
  type?: string;
  canHaveChildren?: boolean;
  metadata?: M;
}

type ComputeTreeRowsFn<P extends TreeParentNode = TreeParentNode> = (parent?: P) => TreeRow[] | Promise<TreeRow[]>;

type TreeStoreOptions = {
  defaultExpandedDepth?: number;
  computeRows: ComputeTreeRowsFn;
}
```


### Example Usage

Here's a simplified example of how this might be used in the Json Schema Viewer use case.

```tsx
// we're storing the original json path on each tree list row, which is helpful in computeSchemaRows func below
type SchemaMetadata = {
  path: JsonPath;
}

// because nobody likes dealing with allOf's :)
const schema = props.mergeAllOf ? mergeAllOf(props.schema) : props.schema;

// does the majority of the work, returns a flat list of TreeRows that are the direct children of the parent
const computeRows: ComputeTreeRowsFn<TreeParentNode<SchemaMetadata>> = (parent) => {
  // if no parent, then we're at level 0 / root of schema
  // if parent, then we should get the appropriate part of our schema, based on parent path
  const parentPath = parent ? parent.metadata.path : [];
  const schemaFragment = parentPath.length ? get(schema, parentPath) : schema;
  
  if (isLocalRef(schemaFragment)) {
    return localRefToSchemaRows(parentPath, schemaFragment.$ref);
  }
  
  if (isRemoteRef(schemaFragment)) {
    return remoteRefToSchemaRows(parentPath, schemaFragment.$ref);
  }

  return computeSchemaRows(parentPath, schemaFragment);
}

const computeSchemaRows = (parentPath: JsonPath, schemaFragment: any): TreeRow[] => {
  const rows: TreeRow[] = [];
  for (const prop in schemaFragment) {
    const n = processNode(parentPath, prop, schemaFragment[prop]);
    if (n) rows.push(n);
  }

  return rows;
}

function processNode(parentPath: JsonPath, propName: JsonPathSegment, node: JSONSchema4): TreeRow | void {
  // note, similar to current processNode in the tree-struct branch, obv need to handle all the different cases
  // this is contrived example
  return {
    id: generateId(),
    name: 'blah',
    type: 'yolo',
    canHaveChildren: foo ? true : false,
    metadata: {
      path: parentPath.concat(propName),
    }
  }
}

const localRefToSchemaRows = (parentPath: JsonPath, ref: string): TreeRow[] => {
  const schemaFragment = get(schema, pointerToPath(ref))
  return computeSchemaRows(parentPath, schemaFragment);
}

const remoteRefToSchemaRows = async (parentPath: JsonPath, ref: string): Promise<TreeRow[]> => {
  const schemaFragment = await getFsOrHttpRefEtc(ref);
  return computeSchemaRows(parentPath, schemaFragment);
}

const store = new TreeStore({
  defaultExpandedDepth: 2,
  computeRows,
});

return <TreeList store={store} />;
```

Issue #71, opened 2020-03-26
Handle local references better


marbemac:
**Describe the bug**

When there are local references, they're not all expandable (not sure if this is array property type problem or not though).

**To Reproduce**

1. Try rendering the schema below with JSV

```yaml
title: Email Message
type: object
x-tags:
  - Email
properties:
  fromAddress:
    $ref: '#/properties/another'
  toAddresses:
    type: array
    items:
      $ref: '#/properties/another'
  ccAddresses:
    type: array
    items:
      $ref: '#/properties/another'
  bccAddresses:
    type: array
    items:
      $ref: '#/properties/another'
  another:
    title: Email
    type: object
    x-tags:
      - Email
    description: An email object
    properties:
      category:
        type: string
      address:
        type: string
```

**Expected behavior**

- [x] `toAddresses`, `ccAddresses`, and `bccAddresses` should be expandable similar to how the `another` property is.
- [x] I don't know that we need to show local $refs (as seen in screenshot below in `fromAddress` property)? @P0lip what do you think?
- [x] Any particular reason for the extra nested object in the `fromAddress` property below? The `another` property is much nicer. cc @P0lip because I think you might have done the original implementation here.
- [x] The array properties `toAddresses`, `ccAddresses`, and `bccAddresses` should have type `array[object]`, or `array[#/propertiesanother]` if we're keeping the local refs showing. One other thought is that we could drop the `array` word there, and go with syntax like `[object]`, `[number]`, etc.

**Screenshots**

<img width="672" alt="Screen Shot 2020-03-25 at 9 19 52 PM" src="https://user-images.githubusercontent.com/847542/77603621-861c0600-6ede-11ea-98c6-d5f96a9c5c9a.png">

P0lip:
>  toAddresses, ccAddresses, and bccAddresses should be expandable similar to how the another property is.

Yeah, this indeed a bug.

>  I don't know that we need to show local $refs (as seen in screenshot below in fromAddress property)? @P0lip what do you think?

I think it may be helpful if you cannot expand ref for whatever reason. In such case, you'd be able to see what the target of your $ref actually is.

> Any particular reason for the extra nested object in the fromAddress property below? The another property is much nicer. cc @P0lip because I think you might have done the original implementation here.

This is needed. Initially we displayed only properties, but this is simply wrong, since we do reference the whole object, not just its properties. This lead to a few subtle bugs, i.e. no validations, etc.

marbemac:
[...]

PR #72, opened 2020-03-31
fix: array local refs

mallachari:
Fixes https://github.com/stoplightio/json-schema-viewer/issues/71

Changes:
- arrays containing ref items can be resolved and expanded

Screenshot from 2020-03-31 13-58-54

### DIFF
--- a/src/components/SchemaRow.tsx
+++ b/src/components/SchemaRow.tsx
@@ -55,7 +55,7 @@ export const SchemaPropertyRow: typeof SchemaRow = ({ node, onGoToRef, rowOption
           style={{
             width: ICON_DIMENSION,
             height: ICON_DIMENSION,
-            ...(isRefNode(schemaNode)
+            ...(isRefNode(schemaNode) && Tree.getLevel(node) === 0
               ? {
                   position: 'relative',
                 }
--- a/src/components/shared/Property.tsx
+++ b/src/components/shared/Property.tsx
@@ -6,7 +6,7 @@ import * as React from 'react';
 import { getSchemaNodeMetadata } from '../../tree/metadata';
 import { GoToRefHandler, IArrayNode, IObjectNode, SchemaKind, SchemaNode, SchemaTreeListNode } from '../../types';
 import { getPrimaryType } from '../../utils/getPrimaryType';
-import { isArrayNodeWithItems, isCombinerNode, isRefNode } from '../../utils/guards';
+import { isArrayNodeWithItems, isCombinerNode, isRefNode, hasRefItems } from '../../utils/guards';
 import { inferType } from '../../utils/inferType';
 import { Types } from './Types';
 
@@ -46,7 +46,7 @@ function isExternalRefSchemaNode(schemaNode: SchemaNode) {
 export const Property: React.FunctionComponent<IProperty> = ({ node: treeNode, onGoToRef }) => {
   const { path, schemaNode: node } = getSchemaNodeMetadata(treeNode);
   const type = isRefNode(node) ? '$ref' : isCombinerNode(node) ? node.combiner : node.type;
-  const subtype = isArrayNodeWithItems(node) ? inferType(node.items) : void 0;
+  const subtype = isArrayNodeWithItems(node) ? (hasRefItems(node) ? '$ref' : inferType(node.items)) : void 0;
 
   const childrenCount = React.useMemo<number | null>(() => {
     if (type === SchemaKind.Object) {
@@ -76,6 +76,7 @@ export const Property: React.FunctionComponent<IProperty> = ({ node: treeNode, o
 
       <Types type={type} subtype={subtype}>
         {isRefNode(node) && node.$ref !== null ? `[${node.$ref}]` : null}
+        {hasRefItems(node) && node.items.$ref !== null ? `[$ref(${node.items.$ref})]` : null}
       </Types>
 
       {onGoToRef && isExternalRefSchemaNode(node) ? (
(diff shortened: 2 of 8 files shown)


src/components/shared/Property.tsx
```diff
@@ -46,7 +46,7 @@ function isExternalRefSchemaNode(schemaNode: SchemaNode) {
 export const Property: React.FunctionComponent<IProperty> = ({ node: treeNode, onGoToRef }) => {
   const { path, schemaNode: node } = getSchemaNodeMetadata(treeNode);
   const type = isRefNode(node) ? '$ref' : isCombinerNode(node) ? node.combiner : node.type;
-  const subtype = isArrayNodeWithItems(node) ? inferType(node.items) : void 0;
+  const subtype = isArrayNodeWithItems(node) ? (hasRefItems(node) ? '$ref' : inferType(node.items)) : void 0;
```
P0lip: Let's just included that check within `inferType` method

src/tree/tree.ts
```diff
@@ -109,22 +112,15 @@ export class SchemaTree extends Tree {
...
-        const refPath = pointerToPath(schemaNode.$ref);
-        const schemaFragment = this.resolveRef
-          ? this.resolveRef(refPath, path, this.schema)
-          : _get(this.schema, refPath);
-        if (!_isObject(schemaFragment)) {
-          throw new ReferenceError(`Could not dereference "${pathToPointer(refPath)}"`);
-        }
-
-        this.populateTreeFragment(node, schemaFragment, path, false);
+      } else if ('$ref' in schemaNode && schemaNode.$ref) {
```
P0lip:
```suggestion
      } else if (isRefNode(schemaNode)) {
```

src/tree/tree.ts
```diff
@@ -109,22 +112,15 @@ export class SchemaTree extends Tree {
...
+        this.populateRefFragment(node, path, schemaNode.$ref);
+      } else if ('items' in schemaNode && schemaNode.items.$ref) {
```
P0lip:
```suggestion
      } else if ('items' in schemaNode && isRefNode(schema.items)) {
```

src/tree/tree.ts
mallachari: `isRefNode` checks only if `$ref` property is in node. It could be still falsy so I would rather make it:
``` } else if (isRefNode(schemaNode) && schemaNode.$ref) {```

or make `isRefNode` check if ref value is there

File: src/components/shared/Property.tsx (hunk 1 above)
Comment (mallachari) [reply src/components/shared/Property.tsx]: Actually that may be not so perfect. `inferType` is used in other places and putting `$ref` there might require to update those parts.
For instance `processNode` in `walk` function uses type that is falsy. We can of course change it to make use of $ref` type but not sure if it's worth it.

File: src/components/SchemaRow.tsx [hunk 4]
```diff
@@ -55,7 +55,7 @@ export const SchemaPropertyRow: typeof SchemaRow = ({ node, onGoToRef, rowOption
           style={{
             width: ICON_DIMENSION,
             height: ICON_DIMENSION,
-            ...(isRefNode(schemaNode)
+            ...(isRefNode(schemaNode) && Tree.getLevel(node) === 0
```
mallachari: @P0lip this refers to last fix for ref nodes caret. I suggest we do that only for level 0 nodes otherwise they may bey shifted

Screenshot from 2020-03-31 16-19-40

P0lip: Ah yes, agreed.

src/components/shared/Property.tsx
P0lip: > Actually that may be not so perfect. inferType is used in other places and putting $ref there might require to update those parts.


Yeah, I'm aware.

> For instance processNode in walk function uses type that is falsy. We can of course change it to make use of $ref` type but not sure if it's worth it.

well, in fact it might require a chunk of work to get it done, so let's leave it as is for now.

src/tree/tree.ts
P0lip: > isRefNode checks only if $ref property is in node. It could be still falsy so I would rather make it:

that's okay if it falsy. We can error.

P0lip:
Could you also add a simple test? We already have a few in place, so should be fairly easy.

marbemac:
🤔 - what do you guys think about the $refs everywhere? Does this actually provide value to a person reading documentation, versus `array[object]` (or even better, pulling title if possible from target ref and using `[]` syntax. For example if "title" at target local ref was `user`, would be `user[]` - if no title there could fall back to `object[]`. This is what we did in the original json-schema-viewer from classic. 

Keep in mind this component is primarily for documentation, not editing use case (hence I think the random local pointers are less valuable and might just add noise). Thoughts?

P0lip:
> thinking - what do you guys think about the $refs everywhere? Does this actually provide value to a person reading documentation, versus array[object] (or even better, pulling title if possible from target ref and using [] syntax. For example if "title" at target local ref was user, would be user[] - if no title there could fall back to object[]. This is what we did in the original json-schema-viewer from classic.

In such case we can simply dereference the whole schema, can't we?

marbemac:
> In such case we can simply dereference the whole schema, can't we?

It's expensive to completely deref gigantic schemas - I was suggesting just reducing visual noise by not displaying all the `#/properties/blah/items/0/foo` type $refs everywhere (under the hood, nothing changes). Because we're bundling by default now, there will be many more random (meaning whatever bundle process optimized to produce, not just original author's local refs) local $refs left in the schema that is passed to json schema viewer.

In any case, we can decide on the above in a separate issue and PR.

---

@P0lip @mallachari where'd we end up here - can you guys work to polish this up and get merged through to v1 studio and platform tomorrow? (and of course, update JSV in v2 at your convenience, but we have some time there).

PR #76, opened 2020-05-22
feat: consume schema title property

P0lip:
General rules:
- given object type & present title, we enrich the type with name, i.e.
```json
{
  "title": "User",
  "type": "object",
  "properties": {
    "name": {
      "type": "string"
    }
  }
}
```
Before:
image

After:
image

- given array type with non-array items, we replace the subtype name with title, i.e.
```json
{
  "type": "array",
  "items": {
    "title": "User",
    "type": "object",
    "properties": {
      "name": {
        "type": "string"
      }
    }
  }
}
```
Before:
image

After:
image

- given array type with no items

```json
{
  "type": "array",
  "title": "Foo"
}
```
Before:
image

After:
image

- $refs - we try to retrieve the title from referenced property

```json
{
  "type": "object",
  "properties": {
    "foo": {
      "$ref": "#/properties/user"
    },
    "user" : {
      "title": "User",
      "properties": {
        "name": {
          "type": "string"
        }
      }
    }
  }
}
```
Before (eager resolving enabled):
image

Before (eager resolving disabled):
image

After (eager resolving enabled):
image

After (eager resolving disabled):
image

```json
{
  "type": "object",
  "properties": {
    "users": {
      "type": "array",
      "items": {
        "$ref": "#/properties/user"
      }
    },
    "user" : {
      "title": "User",
      "properties": {
        "name": {
          "type": "string"
        }
      }
    }
  }
}
```
Before (eager resolving enabled):
image

Before (eager resolving disabled):
image

After (eager resolving enabled):
image
After (eager resolving disabled):
image

Remarks:
- we do not support combiners at all (on purpose)
```
{
  "title": "test",
  "oneOf": [
    { "type": "string" }
  ]
}
```
image

- we do not display types for primitive types (on purpose)

```json
{
  "type": "string",
  "title": "foo"
}
```
image

```json
{
  "type": "array",
  "items": {
    "title": "abc",
    "type": "string"
  }
}
```

- we do not display titles for mixed types (on purpose)
```json
{
  "type": "array",
  "items": {
    "title": "abc",
    "type": ["string", "object"]
  }
}
```
image

- arrays with defined items
```json
{
  "title": "test",
  "type": "array",
  "items": [
    {
      "title": "foo",
      "type": "string"
    },
    {
      "title": "foo",
      "type": "number"
    }
  ]
}
```
image

### DIFF
--- a/src/tree/utils/populateTree.ts
+++ b/src/tree/utils/populateTree.ts
@@ -6,7 +6,7 @@ import { IArrayNode, IObjectNode, SchemaKind, SchemaNode, SchemaTreeListNode } f
 import { getCombiner } from '../../utils/getCombiner';
 import { getPrimaryType } from '../../utils/getPrimaryType';
 import { isCombinerNode, isRefNode } from '../../utils/guards';
-import { getNodeMetadata, metadataStore } from '../metadata';
+import { getNodeMetadata, getSchemaNodeMetadata, metadataStore } from '../metadata';
 import { createErrorTreeNode } from './createErrorTreeNode';
 import { mergeAllOf } from './mergeAllOf';
 import { walk } from './walk';
@@ -51,7 +51,7 @@ export const populateTree: Walker = (schema, parent, level, path, options): unde
     });
 
     if (isRefNode(node) && node.$ref !== null) {
-      (treeNode as TreeListParentNode).children = [];
+      processRef(treeNode, node as JSONSchema4, level, path, options);
     } else if (!isCombinerNode(node)) {
       switch (getPrimaryType(node)) {
         case SchemaKind.Array:
@@ -117,7 +117,7 @@ function processArray(
   }
 
   if ('$ref' in items) {
-    (node as TreeListParentNode).children = [];
+    processRef(node, items, level, path, options);
   } else if (Array.isArray(items)) {
     const children: SchemaTreeListNode[] = [];
     (node as TreeListParentNode).children = children;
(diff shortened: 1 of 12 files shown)


marbemac:
This is 🔥.

One request - if you have not already, could you yalc this into platform v1 and manually inspect a wide variety of the `sample-models` in explorer to make sure that JSV:

a) does not crash for any of them
b) handles the complex use cases represented in that repo nicely

---

The only thing I'm not sure about is `object[User]`. The `[]` kind of throws me off, since it's normally used for arrays. And it's not like when it is an array we show `array[object[User]]`. Maybe we just show `User` rather than `object[User]`?

marbemac:
Also, I suggest that @mallachari does a pass on the code review.

src/tree/utils/populateTree.ts
```diff
@@ -192,6 +192,25 @@ function processObject(
...
+function processRef(
+  node: TreeListNode,
+  schema: JSONSchema4,
+  level: number,
+  path: JsonPath,
+  options: WalkingOptions | null,
+) {
+  (node as TreeListParentNode).children = [];
+  try {
+    const resolved = resolveSchema(schema, path, options);
```
mallachari, line 204: I suppose that won't have any performance impact? It's being resolved elsewhere anyway if I'm correct.

P0lip: Yeah, it shouldn't really have any.

P0lip:
> This might be unrelated here but I also don't understand why is there an extra level of indentation in one of these cases:

$refs have one extra level of nesting unless you enable eager $ref resolving (which we do have enabled in elements - studio, explorer, docs), so it's not that important.

Basically, when you don't resolve $refs upfront, we retain them in the document, to be resolved later once you expand the row.

image

Upon expansion
image

With eager resolving enabled, this is how it looks like
image

It's fairly transparent to user - the rendered schema looks like there was no $ref at all.
Having said that, the extra nesting level is for the sake of consistency.
In our case, however, the effect is rather none, given we have the eager resolving option enabled, so in our case the schema will look like the latter example.

P0lip:
`object - {name}` looks tad weird, but hm, getting rid of `object` may be confusing for some.
Don't have any preference.
@marbemac what do you think about the proposal @mallachari posted?

marbemac:
How about `object{User}` and `array[User]`? In this case would drop the extra `{4}` property counts, not sure how useful those really are in any case.

Or `User` and `User[]`?

P0lip:
> Or User and User[]?

This looks good to me.

P0lip:
I went with image
It feels really cool.
No title: 
image

marbemac:
Looks great!

PR #78, opened 2020-06-01
Fix: safe stringify

mallachari:
Now as `BigInt` numbers may be used we need to stringify them safely.
Though in case of validations we don't need to use stringify as we check for arrays and objects before. Unless there are other possible types that won't work well with `String()`). Are there @P0lip ?
We could just use `safeStringify` but then we won't get wrapping quotes for strings and would need to check for type anyway.

### DIFF
--- a/src/components/shared/Validations.tsx
+++ b/src/components/shared/Validations.tsx
@@ -1,3 +1,4 @@
+import { safeStringify } from '@stoplight/json';
 import { Dictionary } from '@stoplight/types';
 import { Popover } from '@stoplight/ui-kit';
 import cn from 'classnames';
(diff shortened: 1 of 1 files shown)


P0lip:
> Though in case of validations we don't need to use stringify as we check for arrays and objects before. Unless there are other possible types that won't work well with String()). Are there @P0lip ?

Yeah, there might be any type besides fn, including objects and arrays.

src/components/shared/Validations.tsx
```diff
@@ -50,7 +50,7 @@ export const Validations: React.FunctionComponent<IValidations> = ({
                 } else {
                   elem = (
                     <div className="m-1 px-1 bg-gray-2 dark:bg-gray-8 font-bold text-sm rounded" key={index}>
-                      {JSON.stringify(validation)}
+                      {typeof validation === 'string' ? `"${validation}"` : String(validation)}
```
P0lip, line 53:
```suggestion
                      {typeof validation === 'string' ? `"${validation}"` : safeStringify(validation)}
```

PR #80, opened 2020-06-25
fix: render required validation corectly for array[object]

P0lip:


wmhilton:
Tests pass, no bitcoin miner.

PR #81, opened 2020-07-01
fix: eager resolving errors

mallachari:
Related to https://github.com/stoplightio/platform-internal/issues/3007

Resolving function may throw error which should be handled when resolving schema eagerly. 
Instead of replacing all children from parent node with an error one it should add proper child containing that error node.

**Before**:
Peek 2020-07-01 14-28

**After**:
Peek 2020-07-01 14-29

### DIFF
--- a/src/tree/utils/populateTree.ts
+++ b/src/tree/utils/populateTree.ts
@@ -3,13 +3,14 @@ import { JsonPath, Optional } from '@stoplight/types';
 import { JSONSchema4 } from 'json-schema';
 import { isObject as _isObject } from 'lodash';
 import { IArrayNode, IObjectNode, SchemaKind, SchemaNode, SchemaTreeListNode } from '../../types';
+import { generateId } from '../../utils/generateId';
 import { getCombiner } from '../../utils/getCombiner';
 import { getPrimaryType } from '../../utils/getPrimaryType';
 import { isCombinerNode, isRefNode } from '../../utils/guards';
 import { getNodeMetadata, getSchemaNodeMetadata, metadataStore } from '../metadata';
 import { createErrorTreeNode } from './createErrorTreeNode';
 import { mergeAllOf } from './mergeAllOf';
-import { walk } from './walk';
+import { processNode, walk } from './walk';
 
 export type WalkerRefResolver = (path: JsonPath | null, $ref: string) => JSONSchema4;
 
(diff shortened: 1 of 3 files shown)


src/tree/utils/populateTree.ts
```diff
@@ -245,7 +246,22 @@ function prepareSchema(
...
+    (node as TreeListParentNode).children.push(treeNode);
+    metadataStore.set(treeNode, {
+      schemaNode: processNode(schema || {}),
+      schema: schema || {},
+      path,
+    });
+
+    treeNode.children.push(createErrorTreeNode(treeNode, ex.message));
+
+    return null;
```
marcelltoth, line 265: You used to return `undefined`, but now it returns `null`. I guess this was not intended.

By the way this return statement is a **lot** nicer then the `return void doStuff()` we had before. 👍

marcelltoth: I guess this is one of those methods that return `T | null | undefined` where really it could get away with either `T | null` or - preferably - with `T | undefined`. If it is the case, let's narrow it down while we're here, it helps every future reader / user IMO.

src/tree/utils/populateTree.ts
```diff
@@ -245,7 +246,22 @@ function prepareSchema(
...
+    };
+
+    (node as TreeListParentNode).children.push(treeNode);
+    metadataStore.set(treeNode, {
+      schemaNode: processNode(schema || {}),
+      schema: schema || {},
+      path,
+    });
+
+    treeNode.children.push(createErrorTreeNode(treeNode, ex.message));
```
marcelltoth, line 263: Can this go right into the `treeNode` declaration? (as in `children: [createEr....]`)

Or does the `metadataStore.set` call mutate `treeNode`?

src/tree/utils/walk.ts
```diff
@@ -29,7 +29,7 @@ function assignNodeSpecificFields(base: IBaseNode, node: JSONSchema4) {
   }
 }
 
-function processNode(node: JSONSchema4): SchemaNode | void {
```
marcelltoth, line 32: 👏

src/tree/utils/populateTree.ts
mallachari: Somehow I like this `return void (call something)` thing, though never used it so far.

I used `null` as that's the return type of this function and it is still fine with the object checking in `populateTree`.  Previous `undefined` didn't match it. We use `JSONSchema4 | null` in few other places so I would keep it for now.

src/tree/utils/populateTree.ts
mallachari: That might not work as `createErrorTreeNode` extracts metadata from parent which is `treeNode`. So it needs to have that metadata set first.

marcelltoth: Oh I see. Not obvious but makes sense. 👍

src/tree/utils/populateTree.ts
marcelltoth: Just curious why do you prefer it over the much more explicit:
```ts
doSomething();
return;
```
?

Anyway, I still don't like that something can return both `null` **or** `undefined` but I see that's how the whole infrastructure is built up here. This is question out of scope for this PR then, go ahead and merge it.

PR #82, opened 2020-10-01
feat: support multiple combiners

P0lip:
Closes https://github.com/stoplightio/platform-internal/issues/4440

### DIFF
--- a/src/tree/utils/mergeOneOrAnyOf.ts
+++ b/src/tree/utils/mergeOneOrAnyOf.ts
@@ -1,11 +1,13 @@
 import { JSONSchema4 } from 'json-schema';
 
 export function mergeOneOrAnyOf(schema: JSONSchema4, combiner: 'oneOf' | 'anyOf'): JSONSchema4[] {
-  const merged: JSONSchema4[] = [];
-
-  if (!Array.isArray(schema.allOf)) return merged;
-
   const items = schema[combiner];
+
+  if (!Array.isArray(items)) return []; // just in case
+
+  if (!Array.isArray(schema.allOf)) return items;
+
+  const merged: JSONSchema4[] = [];
 
   if (Array.isArray(items)) {
     for (const item of items) {
--- a/src/tree/utils/mergeOneOrAnyOf.ts
+++ b/src/tree/utils/mergeOneOrAnyOf.ts
@@ -1,0 +1,19 @@
+import { JSONSchema4 } from 'json-schema';
+
+export function mergeOneOrAnyOf(schema: JSONSchema4, combiner: 'oneOf' | 'anyOf'): JSONSchema4[] {
+  const merged: JSONSchema4[] = [];
+
+  if (!Array.isArray(schema.allOf)) return merged;
+
+  const items = schema[combiner];
+
+  if (Array.isArray(items)) {
+    for (const item of items) {
+      merged.push({
+        allOf: [item, ...schema.allOf],
+      });
+    }
+  }
+
+  return merged;
+}
--- a/src/tree/utils/mergeOneOrAnyOf.ts
+++ b/src/tree/utils/mergeOneOrAnyOf.ts
@@ -10,7 +10,7 @@ export function mergeOneOrAnyOf(schema: JSONSchema4, combiner: 'oneOf' | 'anyOf'
   if (Array.isArray(items)) {
     for (const item of items) {
       merged.push({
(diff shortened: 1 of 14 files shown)


src/tree/utils/mergeOneOrAnyOf.ts
```diff
@@ -0,0 +1,19 @@
...
+  const merged: JSONSchema4[] = [];
+
+  if (!Array.isArray(schema.allOf)) return merged;
+
+  const items = schema[combiner];
+
+  if (Array.isArray(items)) {
+    for (const item of items) {
+      merged.push({
+        allOf: [item, ...schema.allOf],
```
marbemac, line 13: I suggest putting the `allOf` properties first, since more often than not things like `id` and other common props will be in it. Generally it's nice to show those at the top.

For example, here is redoc - they append the allOf and thus props like `id` are found in the middle of the schema, rather than at the top. Having `id` at top would be nicer IMHO.

Screen Shot 2020-10-02 at 6 35 55 PM

src/tree/utils/populateTree.ts
```diff
@@ -76,6 +82,13 @@ export const populateTree: Walker = (schema, parent, level, path, options): unde
     } else if (_isObject(node.properties)) {
       (treeNode as TreeListParentNode).children = [];
 
+      if (options?.mergeAllOf && (node.combiner === 'oneOf' || node.combiner === 'anyOf')) {
```
marbemac, line 85: Curious - when do we ever not merge all of? Is there a use case in our ecosystem?

src/tree/utils/populateTree.ts
```diff
@@ -76,6 +82,13 @@ export const populateTree: Walker = (schema, parent, level, path, options): unde
...
+        const merged = mergeOneOrAnyOf(fragment, node.combiner);
+        if (merged.length > 0) {
```
marbemac, line 87: What is this guard protecting? If merged is empty, should we not set properties to empty?

src/tree/utils/walk.ts
```diff
@@ -59,36 +61,30 @@ export function processNode(node: JSONSchema4): SchemaNode {
...
-  if ('$ref' in node) {
-    return {
+  } else if ('$ref' in node) {
+    yield {
       id: generateId(),
       $ref: typeof node.$ref !== 'string' ? null : node.$ref,
       ...title,
     };
+  } else if ('not' in node) {
+    // todo: shall we support it?
```
marbemac, line 80: Later 🤓

marbemac (changes requested):
Could you also yalc this into studio and manually verify things work as expected with the complex amadeus models and we didn't break anything?

src/tree/utils/populateTree.ts
P0lip: >  Is there a use case in our ecosystem?

I don't think so. IIRC we didn't want allOf for masking back in the day, but that's no longer relevant.

src/tree/utils/mergeOneOrAnyOf.ts
P0lip: Makes sense! Will do.

src/tree/utils/walk.ts
P0lip: Yeah, it's an old one :laughing:

src/tree/utils/populateTree.ts
P0lip: `mergeOneOrAnyOf` will return an empty array if there is no `allOf`, hence the guard.
I'll tweak `mergeOneOrAnyOf` to make sure we always return `node.properties` if there is no `allOf`.

P0lip:
> Could you also yalc this into studio and manually verify things work as expected with the complex amadeus models and we didn't break anything?

It doesn't seem like anything is broken.

src/tree/utils/populateTree.ts
marbemac: Oh I'm saying that is there ever a reason to leave allOf in place? Versus merging it out.

marbemac (approved):

src/tree/utils/populateTree.ts
P0lip: Yeah, I missed one word in the sentence above and the entire sense was changed.
I said we hadn't wanted allOf merging for masking some time ago, but now, allOf merging is supported by masking and enabled by default.
That being said, there doesn't seem to be any use case anymore.

PR #84, opened 2020-10-29
feat: display format next to type

mmiask:
This PR applies changes already introduced in https://github.com/stoplightio/elements/pull/666 and is a part of https://github.com/stoplightio/elements/issues/653

* Removes `format` from validations displayed on the right side of the row and moves it next to a displayed `type`
* If `type` value is a single type, then the color of the format text is inherited from the type
* If `type` value is an array, it has a default font color (black)

Schema for testing:

```
{
  "title": "model-with-formats",
  "type": "object",
  "properties": {
    "date-of-birth": {
      "type": [
        "number",
        "string",
        "array"
      ],
      "format": "date-time",
      "items": {}
    },
    "name": {
      "type": "string"
    },
    "id": {
      "type": "number",
      "format": "float"
    },
    "notype": {
        "format": "date-time"
    },
    "permissions": {
      "type": [
        "string",
        "object"
      ],
      "format": "password",
      "properties": {
        "ids": {
          "type": "array",
          "items": {
            "type": "integer",
            "format": "int32"
          }
        }
      }
    }
  }
}
```


src/components/shared/Validations.tsx
```diff
@@ -73,3 +77,15 @@ export const Validations: React.FunctionComponent<IValidations> = ({
     </>
   );
 };
+
+export const Format: React.FunctionComponent<{ schemaNode: JSONSchema4 }> = ({ schemaNode }) => {
+  return (
+    <div
+      {...(typeof schemaNode.type === 'string'
```
P0lip, line 84:
```suggestion
      {...(typeof schemaNode.type === 'string' && schemaNode.type in PropertyTypeColors
```

src/components/shared/Validations.tsx
```diff
@@ -73,3 +77,15 @@ export const Validations: React.FunctionComponent<IValidations> = ({
...
+        ? { className: 'ml-2 ' + PropertyTypeColors[schemaNode.type as string] }
```
P0lip, line 85:
```suggestion
        ? { className: `ml-2 ${PropertyTypeColors[schemaNode.type as string]}` }
```
one could also use classnames, IIRC it's already used in a few spots

src/components/shared/Validations.tsx
```diff
@@ -13,7 +15,7 @@ export const Validations: React.FunctionComponent<IValidations> = ({
   required,
   validations: { deprecated, ...validations },
 }) => {
-  const validationCount = Object.keys(validations).length;
+  const validationCount = Object.keys(validations).filter(validation => validation !== 'format').length;
```
P0lip, line 18: Let's just skip `format` in validations entirely.

src/components/SchemaRow.tsx
```diff
@@ -90,6 +90,7 @@ export const SchemaPropertyRow: typeof SchemaRow = ({ node, onGoToRef, rowOption
 
       <div className="flex-1 flex truncate">
         <Property node={node} onGoToRef={onGoToRef} />
+        {metadata.schema.type && metadata.schema.format && <Format schemaNode={metadata.schema} />}
```
P0lip, line 93: and let's do a minor type check, as format can be anything.
```suggestion
        {metadata.schema.type && typeof metadata.schema.format === 'string' && <Format schemaNode={metadata.schema} />}
```

src/components/shared/Validations.tsx
mmiask: I skip the `format` in lines 36-38. Is that what you had in mind?

P0lip: Ah, sorry. I wasn't clear enough.
I meant skipping them in `src/utils/getValidations.ts`

P0lip: https://github.com/stoplightio/json-schema-viewer/blob/f9d11d6a54de347ff283a9bd9f5ea5e729dd0441/src/utils/getValidations.ts#L7

src/components/shared/Validations.tsx
```diff
@@ -14,8 +16,8 @@ export const Validations: React.FunctionComponent<IValidations> = ({
   required,
   validations: { deprecated, readOnly, writeOnly, ...validations },
 }) => {
+  const validationCount = Object.keys(validations).filter(validation => validation !== 'format').length;
```
P0lip, line 19: Alternatively, let's do
```suggestion
const filteredValidations = validations.filter(validation => validation !== 'format');
const validationCount = Object.keys(filteredValidations).length; 
```

src/components/shared/Validations.tsx
```diff
@@ -44,38 +46,40 @@ export const Validations: React.FunctionComponent<IValidations> = ({
           interactionKind="hover"
           content={
             <div className="p-5" style={{ maxHeight: 500, maxWidth: 400 }}>
-              {Object.keys(validations).map((key, index) => {
-                const validation = validations[key];
+              {Object.keys(validations)
+                .filter(validation => validation !== 'format')
+                .map((key, index) => {
```
P0lip, line 51:
```suggestion
              {Object.keys(filteredValidations).map((key, index) => {
```

src/components/shared/Validations.tsx
```diff
@@ -14,8 +16,8 @@ export const Validations: React.FunctionComponent<IValidations> = ({
   required,
   validations: { deprecated, readOnly, writeOnly, ...validations },
 }) => {
-  const viewMode = React.useContext(ViewModeContext);
   const validationCount = Object.keys(validations).length;
+  const viewMode = React.useContext(ViewModeContext);
```
P0lip, line 20:
```suggestion
  const viewMode = React.useContext(ViewModeContext);
  const validationCount = Object.keys(validations).length;
```

P0lip (approved):

src/components/shared/Validations.tsx
```diff
@@ -86,3 +88,15 @@ export const Validations: React.FunctionComponent<IValidations> = ({
     </>
   );
 };
+
+export const Format: React.FunctionComponent<{ schemaNode: JSONSchema4 }> = ({ schemaNode }) => {
```
P0lip, line 92: schemaNode !== JSONSchema4

P0lip (changes requested):
Missed this one

src/components/shared/Validations.tsx
```diff
@@ -2,8 +2,10 @@ import { safeStringify } from '@stoplight/json';
 import { Dictionary } from '@stoplight/types';
 import { Popover } from '@stoplight/ui-kit';
 import cn from 'classnames';
+import { JSONSchema4 } from 'json-schema';
 import * as React from 'react';
 import { ViewModeContext } from '../JsonSchemaViewer';
+import { PropertyTypeColors } from './Types';
```
P0lip, line 8:
```suggestion
import { PropertyTypeColors, SchemaNode } from './Types';
```

P0lip, line 8:
```suggestion
import { PropertyTypeColors } from './Types';
import { SchemaNode } from '../types';
```

src/components/shared/Validations.tsx
```diff
@@ -2,8 +2,10 @@ import { safeStringify } from '@stoplight/json';
 import { Dictionary } from '@stoplight/types';
 import { Popover } from '@stoplight/ui-kit';
 import cn from 'classnames';
+import { JSONSchema4 } from 'json-schema';
```
P0lip, line 5:
```suggestion
```

src/components/SchemaRow.tsx
```diff
@@ -90,6 +90,7 @@ export const SchemaPropertyRow: typeof SchemaRow = ({ node, onGoToRef, rowOption
 
       <div className="flex-1 flex truncate">
         <Property node={node} onGoToRef={onGoToRef} />
+        {metadata.schema.type && typeof metadata.schema.format === 'string' && <Format schemaNode={metadata.schema} />}
```
P0lip, line 93:
```suggestion
        {metadata.schema.type && typeof metadata.schema.format === 'string' && <Format schema={metadata.schema} />}
```

P0lip: Let's use `schemaNode` if possible.

src/components/shared/Validations.tsx
mmiask: My fault, this parameter should actually be named `nodeSchema`, since it accepts a schema of a node. `SchemaNode` probably doesn't not apply here though, beacuse it only contains 2 elements: `id` and `?title` and we are looking for a validation

src/components/shared/Validations.tsx
mmiask: I ended separating `format` from `validations` like this:

```
validations: { deprecated, readOnly, writeOnly, format, ...validations }
```

src/components/shared/Validations.tsx
P0lip: let's call the prop `schema` then.
I think that's how we usually call it.

`SchemaNode` does have more properties. Would only need to include `format` (could be done in a similar way to `enum`).
In any case, using `schema` here is alright.

src/components/SchemaRow.tsx
```diff
@@ -90,6 +90,7 @@ export const SchemaPropertyRow: typeof SchemaRow = ({ node, onGoToRef, rowOption
 
       <div className="flex-1 flex truncate">
         <Property node={node} onGoToRef={onGoToRef} />
+        {metadata.schema.type && typeof metadata.schema.format === 'string' && <Format nodeSchema={metadata.schema} />}
```
P0lip, line 93: Let's use `schema` then. IIRC we use it in other spots when we pass schema around.

P0lip (approved):

PR #86, opened 2020-11-04
fix: handle combiners placed under arrays 

P0lip:
Fixes https://github.com/stoplightio/platform-internal/issues/4527 (do note that rendering combiner next to type is still not supported)

Before:
image
image

After:
image
image


wmhilton (approved):
👍

PR #87, opened 2020-11-04
feat: remove optional tag

mpodlasin:
Relates to https://github.com/stoplightio/elements/issues/655

### DIFF
--- a/src/components/shared/Validations.tsx
+++ b/src/components/shared/Validations.tsx
@@ -1,7 +1,6 @@
 import { safeStringify } from '@stoplight/json';
 import { Dictionary } from '@stoplight/types';
 import { Popover } from '@stoplight/ui-kit';
-import cn from 'classnames';
 import * as React from 'react';
 import { ViewModeContext } from '../JsonSchemaViewer';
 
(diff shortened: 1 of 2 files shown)


src/components/shared/Validations.tsx
```diff
@@ -18,9 +17,9 @@ export const Validations: React.FunctionComponent<IValidations> = ({
   const validationCount = Object.keys(validations).length;
 
   const requiredElem = (
-    <div className={cn('ml-2', required ? 'font-medium' : 'text-darken-7 dark:text-lighten-6')}>
-      {required ? 'required' : 'optional'}
-      {validationCount ? `+${validationCount}` : ''}
+    <div>
+      {required && <span className="ml-2 text-orange-7 dark:text-lighten-6">required</span>}
```
P0lip, line 21: image
```suggestion
      {required && <span className="ml-2 text-orange-7 dark:text-orange-6">required</span>}
```

P0lip (changes requested):
Let's tweak the color a bit.

src/components/shared/Validations.tsx
mpodlasin: uups, didn't notice that.

Fixed, thanks for catching it :)

mmiask:
I know it's probably out of scope for this PR, but we might want to come up with a better solution for displaying validations when they are not directly rendered on the right side of the row

e.g.

Screenshot 2020-11-04 at 13 30 43
 
For `email`,  this `+3` looks a little weird to me, when there are no validations displayed directly on the row. I would suggest rendering either name of a first validation + number of other validations e.g. `min-length + 4` or `validations: count`

philsturgeon:
Right on @mmiask, we've planned a refresh for this area which is being handled by this issue: https://github.com/stoplightio/elements/issues/668

P0lip (approved):

PR #88, opened 2020-11-05
feat: better support of anyOf/oneOf + type scenario

P0lip:
Requires https://github.com/stoplightio/json-schema-viewer/pull/86

Closes_ https://github.com/stoplightio/platform-internal/issues/4382

### DIFF
--- a/src/__fixtures__/allof-with-type.json
+++ b/src/__fixtures__/allof-with-type.json
@@ -1,0 +1,81 @@
+{
+  "title": "BugExample",
+  "description": "An example model to demonstrate a bug.",
+  "allOf": [
+    {
+      "properties": {
+        "actionType": {
+          "type": "string",
+          "enum": [
+            "Cancel",
+            "Confirm",
+            "Create",
+            "Update"
+          ]
+        },
+        "id": {
+          "type": "string",
+          "description": "The identifier of the existing reservation."
+        },
+        "externalId": {
+          "type": "string"
+        },
+        "calculateCosts": {
+          "type": "boolean"
+        },
+        "calculateDates": {
+          "type": "boolean"
+        },
+        "items": {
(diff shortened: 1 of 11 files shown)


P0lip:
@marbemac this is very likely to break masking.
Shall we hide it behind a flag or something? Is displaying an old view acceptable for masking view?

marbemac:
> this is very likely to break masking.

Hmm, what's the example(s) you're thinking of re what might break?

P0lip:
Well, I'm confident the oneOf/anyOf + type scenario will be almost always broken, as property paths will be mismatched.
We'd need to perform the same kind of merging on the backend side of masking for the paths to match.

P0lip:
I kicked of another project called [@stoplight/json-schema-tree](https://github.com/stoplightio/json-schema-tree) that is supposed to take out of all the schema processing logic we have in JSV - do note that although the logic is more or less similar, it's been vastly refactored to payback the debt we have in JSV.
We'll be able to share it between masking & jsv, and possibly JSE.
This will significantly simplify JSV making contributions much easier, in particular when it comes to plain UI changes.

P0lip:
@stoplightio/void-crew anyone? it's ready.

marbemac (approved):
Nice, looks great 👏 . I think the combiner + type merging is specific enough that we roll the dice and hope that design lib folks don't use it much (probably not at all tbh, not many folks are using design libs and masking yet).

wmhilton:
> @stoplightio/void-crew anyone? it's ready.

What do you mean by "ready" exactly? Is it still going to break masking?

P0lip:
Yeah, it might break it at times but what Marc said. It's a rare scenario. Currently we don't display anything in such case so masking doesn't really work anyway.

PR #92, opened 2020-11-14
test: assert HTML output

P0lip:
We assert the structure of the whole jsv component to make sure we don't regress when we switch over to json-schema-tree.
I decided to strip certain attributes as well as remove nodes that we don't care about to make sure to output is as lean as possible.


domagojk:
I think these two cases result in same the snapshot:

```typescript
test('given allOf merging disabled, should preserve both combiners', () => {
    expect(dumpDom(<JsonSchemaViewer schema={schema} expanded={true} />)).toMatchSnapshot();
});

test('given allOf merging enabled, should merge contents of allOf combiners', () => {
   expect(dumpDom(<JsonSchemaViewer schema={schema} expanded={true} mergeAllOf />)).toMatchSnapshot();
});
```

because from what I see `mergeAllOf` is `true` by default?

P0lip:
@domagojk Whooops, yeah, good find.

domagojk (approved):
Even though I don't understand it all, tests look good to me! :)

PR #93, opened 2020-11-24
feat: handle anyof/oneof with selects

mpodlasin:
After recent discussion, we are putting this one on hold.

It however seems to work.

I also managed to make the tests pass, but it should be double-checked if snapshots really look as they should.


marcelltoth:
Closing in favor or #110

PR #96, opened 2020-12-15
chore: remove top bar

P0lip:
Removing top bar nobody has ever used.
This is technically a breaking change, but not from our perspective, since we've never used it.
It's an old next thingy we ported over.

image

`style` prop was never consumed. It was there, but did nothing.


domagojk (approved):
Always nice to see stuff get deleted 😄 
It looks good! 👍

PR #91, opened 2020-11-14
feat: use @stoplight/json-schema-tree 

P0lip:
Complementary PRs: https://github.com/stoplightio/json-schema-viewer/pull/96 https://github.com/stoplightio/json-schema-viewer/pull/97 https://github.com/stoplightio/json-schema-viewer/pull/98 https://github.com/stoplightio/json-schema-viewer/pull/95 #100 

There is one major breaking change from component's perspective - $refs are resolved eagerly by default and there is no way to alter this behavior.
We had that option enabled at all times in all projects anyway, thus it only added redundant complexity for us.
That said, a few tests had to be updated to account for that change.
There was a number of tests updated as a result of better handling of a particular scenario.

image

### $refs

image
image
image

### Before and after

**3.0**
image

**This PR**
image

---

**3.0**
image

**This PR**
image

---

**3.0***
image

**This PR**
image

---

**3.0**
image

**This PR**
image

---

The array flattening logic adds a tiny bit of complexity, but I decided to implement it so that we can leverage this updated component in existing products. I'm aware that it's likely to be thrown away in the next iteration of JSV, but until then, we couldn't benefit from the improvements.

I decided to tweak the naming as well - hopefully I did a good job here :laughing: 

## Changelog

Full list of notable differences vs existing 3.0 (certain pieces were extracted to separate PRs, but listing them here to keep track of them):

- enforced $ref resolving 
- improved flattening - reduced nesting in some case (see array + combiner scenario I attached)
- better oneOf/anyOf merging
- alert icon for $refs errors (see screenshot)
- more correct properties count (see the User in above screenshot, I decided to display the number of actual rows) + more cases covered
- top bar is gone #96 - never used, Next heritage
- better usage of titles (https://github.com/stoplightio/json-schema-viewer/pull/97 + this PR - see Property.spec.tsx)
- format is displayed even for type-less nodes, i.e. `{ format: 'date-time' }`
- ordered array items are handled more properly (`{ "type": "array", "items": [item1, item2, ...] }`)
- Studio bonus -> JST emits error event when all of merging fails making https://github.com/stoplightio/platform-internal/issues/4902 a piece of cake

PR #98, opened 2020-12-16
refactor: move Format component out of Validations

P0lip:
Initially implemented in #91.

`span` felt more appropriate here.


wmhilton (approved):
😂

PR #101, opened 2020-12-23
feat: drop expanded prop

P0lip:
This is breaking from a component's perspective, but not from ours - we usually set `expandedDepth` to 1 and do not use `expanded` anywhere.
Furthemore, migrated a few tests I added yesterday to the other branch.

### DIFF
--- a/src/__stories__/JsonSchemaViewer.tsx
+++ b/src/__stories__/JsonSchemaViewer.tsx
@@ -21,7 +21,6 @@ storiesOf('JsonSchemaViewer', module)
     <JsonSchemaViewer
       schema={schema as JSONSchema4}
       defaultExpandedDepth={number('defaultExpandedDepth', 0)}
-      expanded={boolean('expanded', true)}
       onGoToRef={action('onGoToRef')}
       viewMode={select(
         'viewMode',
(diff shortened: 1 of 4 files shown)


wmhilton:
> we usually set

Inspiring a lot of confidence there @P0lip 😬

P0lip:
Well, the target branch is `beta`, thus it's going to be released as beta (3 I think, at least judging by the past releases https://github.com/stoplightio/json-schema-viewer/releases) that nobody uses yet
We can safely merge the PR.

mallachari:
Are you willing to use beta in nearby in nearby future or is it for us to play with?
We want to build new jsv design (https://github.com/stoplightio/json-schema-viewer/pull/103) on beta though need to be sure it can be used.
Btw. we need your acceptance on this pr and any following.

P0lip:
> Are you willing to use beta in nearby in nearby future or is it for us to play with?

Although, there's no plan to use `beta` in short-term, I assume it's going to be in a reasonable state at all times, meaning we can use it if we need to.

marcelltoth:
@mallachari  I hear your worries. The result of your PR is totally "a reasonable state", if not a perfect one, don't worry about it! 😉

marcelltoth (approved):
Good, I really don't like redundant props

PR #103, opened 2021-02-04
feat: new JSV design

mallachari:
Resolves https://github.com/stoplightio/elements/issues/657
New JSE multi-row design replacing ui-kit with mosaic components.

Description and validations are displayed within a row and row height is calculated based on its number.

The naming format of validations is subject of different issue so I just put them the way they are in schema using similar method they are displayed in Elements. Number validations could be moved next to property and type.

Object types are displayed same way they were previously. The final goal is to include a dropdown next to type and allow selecting and displaying proper type but that's out of scope of this issue.

The remaining problem is multiline description and wrapped elements in validations that should make the row grow dynamically depending on their size - though there's separate issue for that.

Screenshot from 2021-02-04 10-52-05


philsturgeon:
Amazing! So this is solving #657 and just happens to do a little bit of #667 and #668 but is leaving them to be finished separately? 

Fantastic, works for me.

mallachari:
> Amazing! So this is solving #657 and just happens to do a little bit of #667 and #668 but is leaving them to be finished separately?

It does do part of #668 as those validations have big impact on way of rendering rows. 
#667 is separate thing (and quite complex one) so I just mentioned here that handling and displaying multitypes will change, but it's not touched here.

philsturgeon:
Ok great, because an end goal for this epic will be to have no more oneOf, anyOf, or OpenAPI keywords showing here at all. 

Keep up the awesome work!

philsturgeon (approved):

src/__stories__/JsonSchemaViewer.tsx
```diff
@@ -48,7 +49,7 @@ storiesOf('JsonSchemaViewer', module)
         <>
           <SchemaRow treeListNode={node} rowOptions={rowOptions} />
           <div className="flex h-full items-center">
-            <Button className="pl-1 mr-1" small minimal icon={<Icon color="grey" iconSize={12} icon="issue" />} />
+            <Button className="pl-1 mr-1" size="sm" appearance="minimal" icon="issue" />
```
marbemac, line 52: Should almost never have to use `className` - prefer `pl` and `mr` props instead please.

src/components/JsonSchemaViewer.tsx
```diff
@@ -118,23 +119,23 @@ export class JsonSchemaViewerComponent extends React.PureComponent<
     }
 
     return (
-      <div className={cn(className, 'JsonSchemaViewer flex flex-col relative')}>
+      <Flex className={cn(className, 'JsonSchemaViewer relative h-full')}>
```
marbemac, line 122: `pos="relative"` and `h="full"` instead of the class names.

src/components/JsonSchemaViewer.tsx
```diff
@@ -118,23 +119,23 @@ export class JsonSchemaViewerComponent extends React.PureComponent<
...
+      </Flex>
     );
   }
 }
 
 const JsonSchemaFallbackComponent: typeof FallbackComponent = ({ error }) => {
   return (
-    <div className="p-4">
+    <Box p={4}>
       <b className="text-danger">Error</b>
```
marbemac, line 136: `<Text as="b" fontWeight="bold" color="danger">Error</Text>`

src/components/SchemaRow.tsx
```diff
@@ -50,49 +52,55 @@ export const SchemaPropertyRow: React.FunctionComponent<Pick<ISchemaRow, 'rowOpt
...
-      </div>
-
-      <Validations
-        required={isPropertyRequired(schemaNode)}
-        deprecated={isRegularNode(schemaNode) && schemaNode.deprecated}
-        validations={
+      </Flex>
+
+      {typeof description === 'string' && description.length > 0 && (
+        <Flex flex={1} my={2} className="truncate">
```
marbemac, line 94: `<Flex textOverflow="truncate"></Flex>`

src/components/SchemaRow.tsx
```diff
@@ -107,21 +115,27 @@ SchemaPropertyRow.displayName = 'JsonSchemaViewer.SchemaPropertyRow';
...
       <TreeListNodeContext.Provider value={treeListNode}>
-        <div className={cn('px-2 flex-1 w-full max-w-full', className)}>
-          <div
-            className="flex items-center text-sm relative"
-            style={{
-              marginLeft: CARET_ICON_BOX_DIMENSION * Tree.getLevel(treeListNode), // offset for spacing
-            }}
+        <Box flex={1} px={2} className={cn('w-full max-w-full', className)}>
+          <Box
+            className="items-center text-sm relative"
```
marbemac, line 127: `<Box alignItems="center" fontSize="sm" pos="relative"></Box>`

src/components/shared/Caret.tsx
```diff
@@ -1,21 +1,21 @@
...
   <span
-    className="absolute flex justify-center cursor-pointer p-1 rounded hover:bg-darken-3"
+    className="absolute flex justify-center cursor-pointer p-1"
     role="button"
     style={style}
   >
     <Icon
       iconSize={size}
-      icon={isExpanded ? 'caret-down' : 'caret-right'}
+      icon={['fas', isExpanded ? 'chevron-down' : 'chevron-right']}
```
marbemac, line 18: `fas` is the default - I suggest omitting it and just passing props similar to before (but with chevron) e.g. `icon={isExpanded ? 'chevron-down' : 'chevron-right'}`

src/components/shared/Description.tsx
```diff
@@ -1,18 +1,19 @@
...
+import { Box } from '@stoplight/mosaic';
 import * as React from 'react';
 
 export const Description: React.FunctionComponent<{ value: string }> = ({ value }) => (
   <Popover
     boundary="window"
     interactionKind="hover"
-    className="ml-2 flex-1 truncate flex items-baseline"
-    target={<div className="text-darken-7 dark:text-lighten-7 w-full truncate">{value}</div>}
+    className="flex-1 truncate flex items-baseline"
```
marbemac, line 10: All of the tailwind classes are now prefixed under `sl-`. If we really have to use them (in this case we do until can replace with new popover), we should start using the prefixed classes because these non prefixed tailwind classes will go away when ui-kit is removed.

src/components/shared/Description.tsx
```diff
@@ -1,18 +1,19 @@
...
+    target={<Box className="text-darken-7 dark:text-lighten-7 w-full truncate">{value}</Box>}
```
marbemac, line 11: Somethig like `<Box textOverflow="truncate" w="full" color="muted"></Box>` - not 100% sure on the color without looking at rendered version - might be a better option.

src/components/shared/Divider.tsx
```diff
@@ -1,11 +1,12 @@
...
 
 import { COMBINER_PRETTY_NAMES } from '../../consts';
 
-export const Divider: React.FunctionComponent<{ kind: SchemaCombinerName }> = ({ kind }) => (
-  <div className="flex items-center w-full absolute" style={{ top: -9, height: 1 }}>
-    <div className="text-darken-7 dark:text-lighten-8 uppercase text-xs pr-2 -ml-4">{COMBINER_PRETTY_NAMES[kind]}</div>
-    <div className="flex-1 bg-darken-5 dark:bg-lighten-5" style={{ height: 1 }} />
-  </div>
+export const Divider: React.FunctionComponent<{ kind?: SchemaCombinerName, style?: React.CSSProperties }> = ({ kind, style }) => (
+  <Flex align="center" className="w-full absolute" style={{ ...style, ...(kind && { top: -9 }), height: 1 }}>
```
marbemac, line 8: `w="full" pos="absolute"`

marbemac (commented):
Not reviewing the entire thing or anything - but just dropped some comments re canonical usage of Mosaic. We should start removing usage of tailwind classes as much as possible - instead opting for the typed props available on most mosaic components. Of course there are still areas we have to use tailwind classes, that's ok - but eventually we're going to want to remove ui-kit which means no non prefixed tailwind classes.

You can find a list of most of the core style props at https://mosaic.vercel.app/docs/style-props.

mallachari:
> Not reviewing the entire thing or anything - but just dropped some comments re canonical usage of Mosaic for the first few files I looked at. We should start removing usage of tailwind classes as much as possible - instead opting for the typed props available on most mosaic components. Of course there are still areas we have to use tailwind classes, that's ok - but eventually we're going to want to remove ui-kit which means no non prefixed tailwind classes.
> 
> You can find a list of most of the core style props at https://mosaic.vercel.app/docs/style-props.

Yup, I'm doing that in my next comit. I just shared first draft so it could be checked.

src/components/SchemaRow.tsx
mallachari: Thanks, I didn't know this property here, I wanted to use `sl-truncate` class

src/components/shared/Validations.tsx
```diff
@@ -1,90 +1,161 @@
...
-  const requiredElem = (
-    <div>
-      {required && <span className="ml-2 text-orange-7 dark:text-orange-6">required</span>}
-      {validationCount ? <span className="ml-1 text-darken-7 dark:text-lighten-6">+{validationCount}</span> : null}
-    </div>
+  const numberValidations = pick(validations, numberValidationNames);
+  const booleanValidations = omit(
+    pickBy(validations, v => ['true', 'false'].includes(String(v))),
+    excludedValidations,
   );
```
marcelltoth, line 36: Why do you do the exclusion out here for booleans, but not for numbers? I prefer the way you do it for booleans here, so the component doesn't have to worry about what to show, but either way is fine.

src/components/shared/Validations.tsx
```diff
@@ -1,90 +1,161 @@
...
-        />
-      ) : (
-        requiredElem
-      )}
-    </>
+        const exclusive =
+          (key === 'minimum' && validations.exclusiveMinimum) || (key === 'maximum' && validations.exclusiveMaximum)
+            ? true
+            : false;
+        const sign = `${key.includes('min') ? '>' : '<'}${exclusive ? '' : '='}`;
```
marcelltoth, line 70: This one is completely optional but there's a bit too much magic going on to my taste here, and some of the magic depends on property names like the definition of `sign`, which might just be a bit too brittle for future maintenance.

As there are only 8 number validation formats defined, maybe defining them one-by-one won't be much longer but more readable.

Something like:
```tsx
// this typing requires `numberValidationNames` to be defined `as const`.
const numberValidationFormatters: Record<typeof numberValidationNames[number], (value: undefined) => string> = {
  minimum: (value) => `>= ${value}`,
  exclusiveMinimum: (value) => `> ${value}`,
  minItems:(value) => `>= ${value} items`,
  minLength:(value) => `>= ${value} characters`,
  maximum: (value) => `<= ${value}`,
  exclusiveMaximum: (value) => `< ${value}`,
  maxItems: (value) => `< ${value} items`,
  maxLength: (value) => `< ${value} characters`,
}

const NumberValidations = ({ validations, className }: { validations: Partial<Record<typeof numberValidationNames[number], unknown>>; className?: string }) => {
  const entries = Object.entries(validations);
  if (!entries.length) {
    return null;
  }
  return (
    <Flex my={2} color="muted">
      <Text fontWeight="light">Allowed values:</Text>
      {entries.map(([key, value]) => numberValidationFormatters[key](value))
        .map((value, i) => (<Text key={i} ml={2} px={1} fontFamily="mono" border rounded="lg" className={className}>{value}</Text>))}
    </Flex>
  )
};
```

This makes it much easier to add new validations or change the copy of existing ones.

src/__stories__/_styles.scss
```diff
@@ -1,5 +1,6 @@
 @import "~@stoplight/tree-list/styles/_tree-list.scss";
-@import "~@stoplight/ui-kit/styles/_ui-kit.scss";
+@import "~@stoplight/mosaic/styles.css";
+@import "~@stoplight/mosaic/themes/default.css";
```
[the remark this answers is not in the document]
P0lip: > JSV doesn't have a CSS file at all? zipper_mouth_face

It may not. In fact, most of our projects have very lean stylesheets, or no stylesheets at all.
The idea was that we rely on Blueprint's styling as much as possible, and use tailwind for stuff like spacing.
Any custom styles we needed were added to ui-kit.

> This is weird, in this case we need to document that all callers should implement the same @imports, don't we @P0lip ?

Yup, true. I think it might be already documented.

package.json
```diff
@@ -46,13 +46,15 @@
...
-    "classnames": "^2.2.6"
+    "classnames": "^2.2.6",
+    "lodash": "^4.17.19"
   },
   "devDependencies": {
     "@rollup/plugin-typescript": "^3.1.1",
     "@sambego/storybook-state": "^1.3.6",
     "@stoplight/eslint-config": "^1.2.0",
     "@stoplight/markdown-viewer": "^4.3.2",
+    "@stoplight/mosaic": "1.0.0-beta.22",
```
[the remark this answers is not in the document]
mallachari: Right, I put it under ui-kit which was there too.

src/__stories__/_styles.scss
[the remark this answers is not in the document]
mallachari: Yup, it does not. It's used internally, documenting might be good.

src/components/shared/Caret.tsx
```diff
@@ -1,22 +1,25 @@
...
+    cursor="pointer"
     role="button"
     style={style}
+    color="muted"
   >
     <Icon
       iconSize={size}
-      icon={isExpanded ? 'caret-down' : 'caret-right'}
-      className="text-darken-9 dark:text-lighten-7"
+      icon={isExpanded ? 'chevron-down' : 'chevron-right'}
```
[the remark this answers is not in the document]
mallachari: I'm not 100% convinced but I see the point and it's reasonable so I've updated it this way :)

src/components/shared/Description.tsx
```diff
@@ -1,18 +1,7 @@
...
-    target={<div className="text-darken-7 dark:text-lighten-7 w-full truncate">{value}</div>}
-    targetClassName="text-darken-7 dark:text-lighten-6 w-full truncate"
-    content={
-      <div className="p-5" style={{ maxHeight: 500, maxWidth: 400 }}>
-        <MarkdownViewer markdown={value} />
-      </div>
-    }
-  />
+  // TODO (JJ): Add mosaic popover showing full description in MarkdownViewer
+  <Box textOverflow="truncate" w="full" color="muted">{value}</Box>
```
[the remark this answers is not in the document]
mallachari: Yeah, that works perfectly!

src/components/shared/Validations.tsx
mallachari: I like your approach. That is definitely more readable. Already updated

mallachari:
@marcelltoth speaking of ui-kit removal, I've noticed that TreeList component is using it as a peer dependency so even though we don't use it anymore we still need to keep that dependency or at least document its requirement.

src/components/shared/Validations.tsx
mallachari: Booleans are simpler case, for number validations we would need to exclude different more names, so I assumed listing number ones will be easier.

src/components/shared/Divider.tsx
```diff
@@ -1,11 +1,12 @@
 import { SchemaCombinerName } from '@stoplight/json-schema-tree';
+import { Box, Flex } from '@stoplight/mosaic';
 import * as React from 'react';
 
 import { COMBINER_PRETTY_NAMES } from '../../consts';
 
-export const Divider: React.FunctionComponent<{ kind: SchemaCombinerName }> = ({ kind }) => (
-  <div className="flex items-center w-full absolute" style={{ top: -9, height: 1 }}>
-    <div className="text-darken-7 dark:text-lighten-8 uppercase text-xs pr-2 -ml-4">{COMBINER_PRETTY_NAMES[kind]}</div>
-    <div className="flex-1 bg-darken-5 dark:bg-lighten-5" style={{ height: 1 }} />
```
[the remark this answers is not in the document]
mallachari: We did. Mosaic has different way of adding dark mode by inverting theme. I'm adding that.

src/components/shared/Caret.tsx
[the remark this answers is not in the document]
marbemac: FWIW the mosaic icon component will not accept arbitrary strings:

Screen Shot 2021-02-08 at 5 27 25 PM

If going the `import {faChevronDown, faChevronRight} from '@fortawesome/free-solid-svg-icons';` route, careful re resulting bundle and tree-shakability. It's very easy to end up with the entire fortawesome iconset, which is huge, in downstream bundles.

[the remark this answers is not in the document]
marbemac: Re dropping bundled icons - we would treat that as a breaking change. And in any case each individual icon is so small there is little incentive to drop icons in the future. A list of the bundled icons is documented in the `Icon` article - https://mosaic.vercel.app/docs/media/icon (we've added a few since the article was created, I'll get it up to date).

src/components/shared/__tests__/Property.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('Property component', () => {
 
     const wrapper = render(schema);
     expect(wrapper.find(Types)).toHaveHTML(
-      '<span class="text-green-6 dark:text-green-4 truncate">array[string]</span>',
+      '<span class="sl-truncate sl-text-muted">array[string]</span>',
```
marcelltoth, line 67: Have I mentioned that I hate snapshot tests yet? :trollface:

mallachari: Constantly. And I agree :)

src/components/shared/Property.tsx
```diff
@@ -34,29 +35,33 @@ export const Property: React.FunctionComponent<IProperty> = ({ onGoToRef }) => {
   return (
     <>
       {schemaNode.subpath.length > 0 && shouldShowPropertyName(schemaNode) && (
```
marcelltoth, line 37: Could you assert `subpath.length > 0` here instead? I know `subpath` and `schemaNode.subpath` are the same, but this is something you need to check for, when looking at this at first I am worried if this is a potential OOB condition

marcelltoth: You didn't write this, I see. Well in that case it's fully optional. 😄

src/components/shared/Types.tsx
```diff
@@ -5,7 +5,7 @@ import {
   SchemaCombinerName,
   SchemaNodeKind,
 } from '@stoplight/json-schema-tree';
-import cn from 'classnames';
+import { Box, Text } from '@stoplight/mosaic';
 import * as React from 'react';
 
 import { PROPERTY_TYPE_COLORS } from '../../consts';
```
marcelltoth, line 11: With your changes, you could eliminate the entire `PROPERTY_TYPE_COLORS` structure.

The only usage is in this file, but we can replace `keyof typeof PTC` with `SchemaNodeKind | SchemaCombinerName | '$ref'`, or hell, even `string` (the method handles any string just fine), and you're done. -20 lines 🚀

src/components/shared/Validations.tsx
```diff
@@ -1,90 +1,170 @@
...
-  const showVisibilityValidations = viewMode === 'standalone' && !!readOnly !== !!writeOnly;
-  const visibility = showVisibilityValidations ? (
-    readOnly ? (
-      <span className="ml-2 text-darken-7 dark:text-lighten-6">read-only</span>
-    ) : (
-      <span className="ml-2 text-darken-7 dark:text-lighten-6">write-only</span>
-    )
-  ) : null;
+// this typing requires `numberValidationNames` to be defined `as const`.
+const numberValidationFormatters: Record<typeof numberValidationNames[number], (value: unknown) => string> = {
```
marcelltoth, line 25: ⚠️  `typeof numberValidationNames[number]` only "works" if you declare
```tsx
const numberValidationNames = [
   // ...
] as const;
```
Otherwise it's a string. I mean, string is fine too if you prefer, but then let's express it as `string` rather than `typeof numberValidationNames[number]` 😄

marcelltoth (approved):
Approved but please check out these comments. You don't need to request another review though unless you're adding some significant changes.

marcelltoth (approved):

src/components/SchemaTree.tsx
```diff
@@ -42,10 +44,24 @@ export const SchemaTree: React.FC<ISchemaTree> = props => {
 
   return (
     <TreeList
+      className="sl-flex-1"
       draggable={false}
-      striped
+      interactive={false}
```
P0lip, line 49: Are you sure about this one?
There is no pointer cursor when you hover over a row.

src/components/shared/Validations.tsx
```diff
@@ -1,90 +1,169 @@
-import { safeStringify } from '@stoplight/json';
-import { Dictionary } from '@stoplight/types';
-import { Popover } from '@stoplight/ui-kit';
+import { RegularNode } from '@stoplight/json-schema-tree';
+import { Flex, Text } from '@stoplight/mosaic';
+import { Dictionary, Primitive } from '@stoplight/types';
+import { capitalize, keys, omit, pick, pickBy, uniq } from 'lodash';
```
P0lip, line 4: `keys` - any reason we can't just use `Object.keys`?

P0lip (approved):

src/components/SchemaTree.tsx
mallachari: True, that was to remove the hover background but probably can be done differently.

src/components/shared/Validations.tsx
mallachari: Only to be consistent with other lodash functions that we needed to use.

src/components/SchemaTree.tsx
mallachari: Actually interactive mode looks good now, so might stay ;)

PR #105, opened 2021-02-15
feat: redesign validations

mallachari:
Resolves https://github.com/stoplightio/elements/issues/668

Makes validations being displayed in more human friendly way. 

image

### DIFF
--- a/src/__fixtures__/default-schema.json
+++ b/src/__fixtures__/default-schema.json
@@ -21,7 +21,7 @@
       "type": "string",
       "format": "date-time",
       "writeOnly": true,
-      "pattern": "MM/dd/yyyy H:mm"
+      "pattern": "^([0-1]?[0-9]|2[0-3]):[0-5][0-9]$"
     },
     "items": {
       "type": ["null", "array"],
@@ -35,7 +35,7 @@
     "email": {
       "type": "string",
       "format": "email",
-      "example": ["one@email.com", "two@email.com"],
+      "examples": ["one@email.com", "two@email.com"],
       "deprecated": true,
       "default": "default@email.com",
       "minLength": 2
--- a/src/__fixtures__/default-schema.json
+++ b/src/__fixtures__/default-schema.json
@@ -4,34 +4,40 @@
   "properties": {
     "name": {
       "type": "string",
+      "const": "Constant name",
+      "examples": ["Example name", "Different name"],
       "description": "The user's full name. This description can be long and should truncate once it reaches the end of the row. If it's not truncating then theres and issue that needs to be fixed. Help!"
     },
     "age": {
       "type": "number",
-      "minimum": 0,
-      "maximum": 150,
+      "minimum": 10,
+      "maximum": 40,
       "multipleOf": 10,
-      "exclusiveMinimum": true,
-      "exclusiveMaximum": true,
+      "default": 20,
+      "enum": [10, 20, 30, 40],
       "readOnly": true
     },
     "completed_at": {
       "type": "string",
       "format": "date-time",
-      "writeOnly": true
+      "writeOnly": true,
+      "pattern": "MM/dd/yyyy H:mm"
     },
     "items": {
       "type": ["null", "array"],
       "items": {
         "type": ["string", "number"]
       },
+      "minItems": 1,
+      "maxItems": 4,
       "description": "This description can be long and should truncate once it reaches the end of the row. If it's not truncating then theres and issue that needs to be fixed. Help!"
     },
     "email": {
       "type": "string",
       "format": "email",
-      "example": "email@email.com",
+      "example": ["one@email.com", "two@email.com"],
       "deprecated": true,
+      "default": "default@email.com",
       "minLength": 2
     },
     "plan": {
--- a/src/components/shared/Validations.tsx
+++ b/src/components/shared/Validations.tsx
@@ -37,26 +37,32 @@ const numberValidationFormatters: Record<string, (value: unknown) => string> = {
   maxLength: value => `<= ${value} characters`,
 };
 
-const validationsFormatter = (name: string) => (value: unknown[] | unknown): ValidationFormat | null => {
+const createStringFormatter = (nowrap: boolean | undefined) => (value: unknown) => {
+  return nowrap || typeof value !== 'string' ? String(value) : `"${value}"`;
+};
+
+const createValidationsFormatter = (name: string, options?: { exact?: boolean; nowrap?: boolean }) => (
+  value: unknown[] | unknown,
+): ValidationFormat | null => {
   const values = Array.isArray(value) ? value : [value];
   if (values.length) {
     return {
-      name: values.length > 1 ? `${name} values` : `${name} value`,
-      values: values.map(stringifyValue),
+      name: options?.exact ? name : values.length > 1 ? `${name} values` : `${name} value`,
+      values: values.map(createStringFormatter(options?.nowrap)),
     };
   }
   return null;
 };
 
 const validationFormatters: Record<string, (value: unknown) => ValidationFormat | null> = {
-  ['const']: validationsFormatter('Allowed'),
-  enum: validationsFormatter('Allowed'),
-  examples: validationsFormatter('Example'),
-  example: validationsFormatter('Example'),
-  ['x-example']: validationsFormatter('Example'),
-  multipleOf: validationsFormatter('Multiple of'),
-  pattern: validationsFormatter('Match pattern'),
-  default: validationsFormatter('Default'),
+  ['const']: createValidationsFormatter('Allowed'),
+  enum: createValidationsFormatter('Allowed'),
+  examples: createValidationsFormatter('Example'),
+  example: createValidationsFormatter('Example'),
+  ['x-example']: createValidationsFormatter('Example'),
+  multipleOf: createValidationsFormatter('Multiple of', { exact: true }),
+  pattern: createValidationsFormatter('Match pattern', { exact: true, nowrap: true }),
+  default: createValidationsFormatter('Default'),
 };
 
 export const Validations: React.FunctionComponent<IValidations> = ({ validations }) => {
@@ -125,7 +131,7 @@ const KeyValueValidation = ({ className, name, values }: { className?: string; n
   return (
     <Flex flexWrap color="muted" my={2} className={className}>
       <Text fontWeight="light">{capitalize(name)}:</Text>
-      {values.map(value => (
+      {uniq(values).map(value => (
         <Text key={value} ml={2} px={1} fontFamily="mono" border rounded="lg" className={className}>
           {value}
         </Text>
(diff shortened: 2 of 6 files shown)


src/components/shared/Validations.tsx
```diff
@@ -28,8 +33,30 @@ const numberValidationFormatters: Record<string, (value: unknown) => string> = {
   minLength: value => `>= ${value} characters`,
   maximum: value => `<= ${value}`,
   exclusiveMaximum: value => `< ${value}`,
-  maxItems: value => `< ${value} items`,
-  maxLength: value => `< ${value} characters`,
+  maxItems: value => `<= ${value} items`,
+  maxLength: value => `<= ${value} characters`,
+};
+
+const validationsFormatter = (name: string) => (value: unknown[] | unknown): ValidationFormat | null => {
```
marcelltoth, line 40: An `unknown[] | unknown` is still `unknown`, although I can see your point that it increases clarity somewhat. 

We shouldn't be using this many `unknown`s to begin with, but that's not your fault, that's how validations are defined unfortunately.

src/components/shared/Validations.tsx
```diff
@@ -28,8 +33,30 @@ const numberValidationFormatters: Record<string, (value: unknown) => string> = {
   minLength: value => `>= ${value} characters`,
   maximum: value => `<= ${value}`,
   exclusiveMaximum: value => `< ${value}`,
-  maxItems: value => `< ${value} items`,
-  maxLength: value => `< ${value} characters`,
+  maxItems: value => `<= ${value} items`,
+  maxLength: value => `<= ${value} characters`,
```
marcelltoth, line 37: Nice catch, my bad!

src/components/shared/Validations.tsx
```diff
@@ -81,48 +107,29 @@ const NumberValidations = ({
...
-    return (
-      <>
-        {keys(value).map(key => (
-          <KeyValueValidation key={key} className={className} name={`${name}.${key}`} value={value[key]} />
-        ))}
-      </>
-    );
-  }
-  const validation = Array.isArray(value) ? value : [value];
+const KeyValueValidation = ({ className, name, values }: { className?: string; name: string; values: string[] }) => {
```
marcelltoth, line 124: This function is much clearer now, great job!

src/components/shared/Validations.tsx
```diff
@@ -81,48 +107,29 @@ const NumberValidations = ({
...
-            <Text key={value} ml={2} px={1} fontFamily="mono" border rounded="lg" className={className}>
-              {value}
-            </Text>
-          );
-        })}
+      {values.map(value => (
+        <Text key={value} ml={2} px={1} fontFamily="mono" border rounded="lg" className={className}>
+          {value}
+        </Text>
+      ))}
```
marcelltoth, line 132: Are you sure `value`s are unique? I know that's how things *generally* are, but I like to be very cautious with `key`s, because they cause hard-to-debug undefined behavior when non-unique. Please either run it through `uniq` or use another key, such as the index. Whichever you think fits the business requirements better.

src/components/shared/__tests__/Validations.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    const validations = getValidationsFromSchema(node);
+    const wrapper = mount(<Validations validations={validations} />);
+
+    expect(wrapper).toIncludeText('>= 10<= 40');
+    expect(wrapper).toIncludeText('Allowed values:10203040');
+    expect(wrapper).toIncludeText('Default value:20');
+    expect(wrapper).toIncludeText('Multiple of value:10');
+    expect(wrapper).toIncludeText('Example value:20');
```
marcelltoth, line 29: I see you're attempting to use the good approach even with the not-so-good framework. We can likely discuss introduction of RTL in here later. Looks good for now.

src/components/shared/Validations.tsx
marcelltoth: Also I'd consider calling the function `createValidationsFormatter` as it's a HOF. It isn't the formatter itself, rather it *creates* the function that will be the formatter.

src/components/shared/Validations.tsx
```diff
@@ -28,8 +33,30 @@ const numberValidationFormatters: Record<string, (value: unknown) => string> = {
...
+};
+
+const validationFormatters: Record<string, (value: unknown) => ValidationFormat | null> = {
+  ['const']: validationsFormatter('Allowed'),
+  enum: validationsFormatter('Allowed'),
+  examples: validationsFormatter('Example'),
+  example: validationsFormatter('Example'),
+  ['x-example']: validationsFormatter('Example'),
+  multipleOf: validationsFormatter('Multiple of'),
+  pattern: validationsFormatter('Match pattern'),
```
marcelltoth, line 58: Your abstraction falls a bit short here. These two shouldn't say `values`.

I'm willing to let it slide for now though if @philsturgeon is, too.

cc @philsturgeon Is it acceptable that it currently says "Match pattern value" and "Multiple of value"?

marcelltoth (changes requested):
Looking good, great work Jakub!

The hard blockers are "match pattern value" because it's a functional non-compliance (unless @philsturgeon says it's OK), and the potential key duplication bug, but that's trivial to fix.

src/components/shared/Validations.tsx
mallachari: Oh yeah, right, it shouldn't. I'll think of better way

src/components/shared/__tests__/Validations.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+      multipleOf: 10,
+      default: 20,
+      example: 20,
+      enum: [10, 20, 30, 40],
+    });
+
+    const validations = getValidationsFromSchema(node);
+    const wrapper = mount(<Validations validations={validations} />);
+
+    expect(wrapper).toIncludeText('>= 10<= 40');
```
P0lip, line 25: You could probably just do
```suggestion
    expect(wrapper.text()).toMatchInlineSnapshot(
```
and verify the output

src/components/shared/Validations.tsx
```diff
@@ -161,9 +168,17 @@ export function validationCount(schemaNode: RegularNode) {
...
+          ...(schemaNode.annotations.examples ? { examples: schemaNode.annotations.examples } : null),
+        }
       : null),
+    ...('fragment' in schemaNode && 'const' in schemaNode.fragment ? { const: schemaNode.fragment.const } : null),
     ...schemaNode.validations,
   };
 }
+
+function stringifyValue(value: unknown) {
+  return typeof value === 'string' ? `"${value}"` : String(value);
```
P0lip: Any reason not to use `JSON.stringify`?

src/components/shared/Validations.tsx
```diff
@@ -28,8 +33,30 @@ const numberValidationFormatters: Record<string, (value: unknown) => string> = {
...
+  }
+  return null;
+};
+
+const validationFormatters: Record<string, (value: unknown) => ValidationFormat | null> = {
+  ['const']: validationsFormatter('Allowed'),
+  enum: validationsFormatter('Allowed'),
+  examples: validationsFormatter('Example'),
+  example: validationsFormatter('Example'),
+  ['x-example']: validationsFormatter('Example'),
```
P0lip: I'm not sure if json-schema-tree supports this. You might want to verify it.

src/components/shared/Validations.tsx
mallachari: Right, might be better

src/components/shared/Validations.tsx
mallachari: It might not. Same with `const` so I took it from `fragment` to not update JST but we might want to do that. Though not sure if `x-example` is worth it.

src/components/shared/__tests__/Validations.spec.tsx
mallachari: Do we like snaphosts? :wink:

marcelltoth: 😭

marcelltoth (approved):

mallachari:
@philsturgeon do you have any remarks from your point of view?

src/components/shared/__tests__/Validations.spec.tsx
P0lip: Well, I don't like them, BUT inline snapshots are okay.

src/components/shared/Validations.tsx
P0lip: >  so I took it from fragment to not update JST

We shouldn't be using `fragment`. :grimacing:

src/components/shared/Validations.tsx
```diff
@@ -161,8 +174,17 @@ export function validationCount(schemaNode: RegularNode) {
...
     ...(schemaNode.enum !== null ? { enum: schemaNode.enum } : null),
-    ...('annotations' in schemaNode && schemaNode.annotations.default
-      ? { default: schemaNode.annotations.default }
+    ...('annotations' in schemaNode
+      ? {
+          ...(schemaNode.annotations.default ? { default: schemaNode.annotations.default } : null),
+          ...(schemaNode.annotations.examples ? { examples: schemaNode.annotations.examples } : null),
+        }
+      : null),
+    ...('fragment' in schemaNode
```
P0lip, line 183: We shouldn't be using fragment.
You can leave a todo note here (or even better create an issue in jst) and I'll get it done soon.

mallachari: Alright, I'll put a note here and create issue for JST

src/__fixtures__/default-schema.json
```diff
@@ -4,34 +4,40 @@
...
+      "default": 20,
+      "enum": [10, 20, 30, 40],
       "readOnly": true
     },
     "completed_at": {
       "type": "string",
       "format": "date-time",
-      "writeOnly": true
+      "writeOnly": true,
+      "pattern": "MM/dd/yyyy H:mm"
```
philsturgeon, line 24: I’m not sure what this is doing but pattern is meant to be regex. 

> pattern (This string SHOULD be a valid regular expression, according to the Ecma-262 Edition 5.1 regular expression dialect)

From OASv3.0.3.

src/__fixtures__/default-schema.json
```diff
@@ -4,34 +4,40 @@
...
       },
+      "minItems": 1,
+      "maxItems": 4,
       "description": "This description can be long and should truncate once it reaches the end of the row. If it's not truncating then theres and issue that needs to be fixed. Help!"
     },
     "email": {
       "type": "string",
       "format": "email",
-      "example": "email@email.com",
+      "example": ["one@email.com", "two@email.com"],
```
philsturgeon, line 38: This is an invalid example. It’s a strong but you’ve got an example of an array that contains two strings.

philsturgeon (commented):
Not sure if these comments matter but I noticed some invalid OAS. I’ll review on computer soon but I’m sick and dying today so still try and look again when I can brain.

src/__fixtures__/default-schema.json
mallachari: Right, changed to regex

src/__fixtures__/default-schema.json
mallachari: I've changed it to `examples` which has to be an array

src/components/shared/Validations.tsx
mallachari: JST bumped and fragment usage removed

P0lip (approved):

PR #109, opened 2021-03-04
refactor: simplify JSV

marcelltoth:
Alright, so this is going to be opinionated. Prep your coffee, tea, monster (😉), whatever. Don't read in a bad mood please. 😇 

## Background

We've long had an issue, stoplightio/elements#667. If you read the description you can see that while it's a non-trivial problem, it *feels* reasonable, something that *can* be done.

Then we tried to implement it. We struggled. Then we invited @P0lip to help us. Then we tried again. Struggled still. Then we held a 4-hour long session together to make something useful. We still struggled, but we built some POC. I tried to build something complete myself based on our POC, but I failed. I was running into all sorts of hard-to-debug state-management-related issues, crashes, etc. Items missing from caches, TypeError-s, etc. I gave up.

### TLDR

While the current state management approach JSV is using has some wonderful features, it is unfortunately borderline impossible to comprehend by basic human beings like myself. This poses a huge risk to the Elements project (and possibly the Stoplight ecosystem as a whole) because - as this oneOf/anyOf issue shows - Team Undefined seems to be simply incapable of working on it, in it's current form.

## Proposal

I propose to remove most of the state-management complexity from JSV, and fall back to a basic, idiomatic react approach:
- No mobx
- No stores of any kind
- No tree-list
- No virtualization
- Much less context usage, much more props passing

The foundation, `json-schema-tree` is kept as-is, JST is an amazing library for abstracting away the bulk of the logic. There wasn't much extra logic actually on top of what JST offers, the most notable one being array element type hoisting. (a.k.a. "flattening") That logic was moved from the _SchemaTreeListTree_ class to custom hooks and functions, see *utils.ts*

The bulk of the presentation logic is kept intact, except it is now simpler as it:
- only uses a single source of input: the JST node, no jumping between a tree list node and a JST node
- uses only props for the most part, not a lot of context, which makes components easier to reason about
- some typing improvements were made, see `PrimitiveArrayNode`, etc

### The good

1. It's *hopefully* **much simpler** to understand and work on. There is evidence of this as I have built a rudimentary version of the oneOf/anyOf selector [in only about 70 LOC](https://github.com/stoplightio/json-schema-viewer/pull/110/files) (and in about 2 hours) which already works much more reliably than our old POC.
2. It unknowingly **fixed some bugs**. Come [here](https://stoplightio.github.io/json-schema-viewer/?path=/story/jsonschemaviewer--default) open *Permissions* then *ids*. It's broken, I found it when comparing the new with the old. The new version just works. It's much more primitive logic, but because of that, there's likely less edge cases and random bugs like this.
3. Removed ~400 LOC, improving readability and bundle size.
3. **Bundle size** is much better. Own size is 12KB vs 20KB prior, in addition to losing two big dependencies:
   - `tree-list`, which is [37KB minified](https://bundlephobia.com/result?p=@stoplight/tree-list@5.0.3)
   - `mobx` & `mobx-react-lite` which are [57 KB](https://bundlephobia.com/result?p=mobx@6.1.8) and [6 KB](https://bundlephobia.com/result?p=mobx-react-lite@3.2.0) respectively.
4. Not having virtualization soles a lot of our current and future problems, such as:
   - Struggling to instantiate JSV into a non-fixed-height container.
   - Not being able to show the entire description if the user asks for it. (Word wrapping and virtualization don't play well together.)
   - Being able to render and assert against JSV in a JSDOM environment without mocking either JSV or autosizer or both.
5. stoplightio/platform-internal#5476 should be implicitly resolved by this change.

### The bad
1. Initial rendering performance is worse for large schemas, especially if `defaultExpandedDepth > 1`. 
  
  The stress test which renders a 15-thousand-line json-schema **twice** with `defaultExpandedDepth: 2` takes a little more than 1 second on my machine vs the virtualized version, which takes about 300ms.
  
  I would say this is still definitely the "worth it" category. The results above are without any attempt at performance optimization, but even if we can't improve it much: 30k lines worth of content loading in 1 sec should be acceptable IMO. We'll throw a loading-spinner, animation or anything and it will feel butter-smooth. (Or a progress-bar, right, @wmhilton? 😉)

## TODO
1. While working on the POC I removed some props, some of which may be needed, I need to add them back.
2. As tests were largely relying on Snapshots and props passing, I need to update and/or rewrite them.


src/tree/types.ts
```diff
@@ -20,10 +20,20 @@ export type SchemaTreeOptions = {
...
+
+export type ComplexArrayNode = ArrayNode & {
+  children: [RegularNode & { simple: false }];
+};
+
+export type BrokenRefArrayNode = ArrayNode & {
+  children: [ReferenceNode & { error: string }];
+};
+
+export type FlattenableNode = PrimitiveArrayNode | ComplexArrayNode | BrokenRefArrayNode;
```
marcelltoth, line 39: If you watch closely, you can see that this is the same as the old `FlattenableNode`, just approached from another angle.

mpodlasin (commented):
For me this is insanely impressive. 

However I am obviously waiting for the passing tests - my main gripe would be catching any regressions, so final opinion on this PR would largely depend for me on quality of test refactor.

mallachari:
I'm surprised that removing tree list went so smoothly. I was expecting much bigger refactor. At this point the code is way more accessible and considering current requirements that change might be good to have. I got used to tree list usage, also got familiar with it to some extent, but trying to add recent changes made it too painful and unnecessarily complex. In the end it's a matter of picking between virtualization and simplicity. 
I didn't notice anything breaking with this approach so far, though porting tests might encounter something that was missed.
Generally I like this approach as it solves problems that we are struggling with for some time, though I'm curious of the opinion from the void side.

P0lip:
> I'm surprised that removing tree list went so smoothly. I was expecting much bigger refactor. At this point the code is way more accessible and considering current requirements that change might be good to have. I got used to tree list usage, also got familiar with it to some extent, but trying to add recent changes made it too painful and unnecessarily complex. In the end it's a matter of picking between virtualization and simplicity.

@mallachari yeah, it's probably because the last time you looked at the component, JSV was strictly tied to tree-list.
Plenty of UI components accessed `tree-list` nodes directly (its `metadata`), and the whole tree was somewhat built on top of tree-list's tree (we simply created tree-list tree nodes when iterating over schemas, etc.)
It was painful as hell to get rid of that `tree-list` usage. I had to spend 2-3 days to do that, but had to do it, so that you all don't hate me :trollface: 
Once JSV was set to use JST, the tree-list was actually used in one spot, that was still quite tangled, but there was not much I could do about it, cause we still had to map JST node with a tree-list node and so on.

---

Speaking of the PR, I'm going to review it soon, probably today or tomorrow.

yarn.lock
```diff
@@ -11028,16 +11017,6 @@ mkdirp@1.x, mkdirp@^1.0.4:
   dependencies:
     minimist "0.0.8"
 
-mobx-react-lite@^1.5.2:
-  version "1.5.2"
-  resolved "https://registry.yarnpkg.com/mobx-react-lite/-/mobx-react-lite-1.5.2.tgz#c4395b0568b9cb16f07669d8869cc4efa1b8656d"
-  integrity sha512-PyZmARqqWtpuQaAoHF5pKX7h6TKNLwq6vtovm4zZvG6sEbMRHHSqioGXSeQbpRmG8Kw8uln3q/W1yMO5IfL5Sg==
-
-mobx@^5.13.0:
```
P0lip, line 11036: Always glad to see this guy gone.
I've always hated the fact `tree-list` used `mobx`, as it was always tad redundant, but since Studio and all other components used it, it never really had any impact on the bundle size.
Now, when we use some other crap that we'll probably abandon in 2022 (:trollface:), it makes sense to drop it.
FWIW, I have a PR in tree-list repo that swaps `mobx` with a tiny custom thingy -> https://github.com/stoplightio/tree-list/blob/dfc024db74b1b7de2f81d50a4d674a3729861344/src/observable.ts.
We've never merged it, but perhaps I get to merge it, so that `RequestMaker` (if it still uses tree-list) does not have that dependency included.

marcelltoth:
Oh, 

> @mallachari yeah, it's probably because the last time you looked at the component, JSV was strictly tied to tree-list.
> Plenty of UI components accessed `tree-list` nodes directly (its `metadata`), and the whole tree was somewhat built on top of tree-list's tree (we simply created tree-list tree nodes when iterating over schemas, etc.)
> It was painful as hell to get rid of that `tree-list` usage. I had to spend 2-3 days to do that, but had to do it, so that you all don't hate me :trollface:
> Once JSV was set to use JST, the tree-list was actually used in one spot, that was still quite tangled, but there was not much I could do about it, cause we still had to map JST node with a tree-list node and so on.

Oh yeah, I wouldn't have attempted this a few months ago 😂 . The JST-extraction was amazing, it took out most of the core logic without having any UI-related dependencies. This is philosophically meant to be a follow-up to that PR, completing what you started there.

yarn.lock
marcelltoth: FYI it's 2021, so `RequestMaker`, in the form you know it, will be gone soon, it's deprecated. 😎

 In the upcoming Elements v7 we have something new that doesn't rely on `tree-list` anymore. So if we eliminate `mobx` here, that should be enough to finally get rid of `mobx` in Elements completely. Doesn't help platform but it helps future opensource users.

src/__stories__/JsonSchemaViewer.tsx
```diff
@@ -37,50 +39,24 @@ storiesOf('JsonSchemaViewer', module)
...
-    );
-  })
   .add('stress-test schema', () => (
     <>
-      <div style={{ height: 345 }}>
+      <div style={{ height: 345, overflowY: 'scroll' }}>
         <JsonSchemaViewer
           schema={stressSchema as JSONSchema4}
           defaultExpandedDepth={number('defaultExpandedDepth', 2)}
-          onGoToRef={action('onGoToRef')}
```
P0lip, line 74: Don't we need that anymore?
It was built for Studio so that you could activate a local source node

src/__stories__/JsonSchemaViewer.tsx
```diff
@@ -37,50 +39,24 @@ storiesOf('JsonSchemaViewer', module)
     <JsonSchemaViewer
       schema={object('schema', {})}
       defaultExpandedDepth={number('defaultExpandedDepth', 0)}
-      onGoToRef={action('onGoToRef')}
       maxRows={number('maxRows', 5)}
       mergeAllOf={boolean('mergeAllOf', true)}
     />
   ))
-  .add('custom row renderer', () => {
```
P0lip, line 45: Ditto.
What about masking? Masking made use of it.

src/contexts/jsvOptions.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+
+type JSVOptions = {
+  defaultExpandedDepth: number;
+  maxRows: number;
```
P0lip, line 5: Thinking about this.
Does it make any sense to keep this?
This was mostly useful for instrumenting tree-list to render X amount of rows at the same time.
Now, you should be able to control it with plain styles.

yarn.lock
P0lip: > Doesn't help platform but it helps future opensource users.

:100:

src/contexts/jsvOptions.tsx
marcelltoth: Absolutely not, it doesn't even work. I forgot to remove it, will do. :D

src/__stories__/JsonSchemaViewer.tsx
marcelltoth: Got it, I'll add it back tomorrow. See TODO item 1.

src/components/SchemaRow.tsx
```diff
@@ -1,128 +1,116 @@
...
 import { Properties } from './shared/Properties';
 
-export interface ISchemaRow {
-  className?: string;
-  treeListNode: SchemaTreeListNode;
-  rowOptions: IRowRendererOptions;
-  onGoToRef?: GoToRefHandler;
+export interface SchemaRowProps {
+  schemaNode: SchemaNode;
+  nestingLevel: number;
```
P0lip, line 22: I could perhaps make it a part of `SchemaNode`.
This is likely to be useful overall.

src/__fixtures__/arrays/of-objects.json
```diff
@@ -8,6 +8,7 @@
       "type": ["array"],
       "items": {
         "type": "object",
+        "title": "SomeObj",
```
P0lip, line 11:
```suggestion
```
unintended?

src/components/SchemaRow.tsx
P0lip: Ah, actually, there's a property called `depth`.
https://github.com/stoplightio/json-schema-tree/blob/beta/src/nodes/BaseNode.ts#L19
Did you run into issues when trying to use this?

P0lip:
I'm okay with dropping tree list, therefore I'm also fine with the chosen approach
Ping me when this PR is ready and I'll be happy to have a final pass.

domagojk:
The PR looks great! So much more deletions than insertions  🤩  But I haven't look into detail, I just wanted to say that this PR will automatically close this one:

Screen Shot 2021-03-10 at 16 46 18

And I guess we shouldn't do this because we still need to update it in studio 🤷‍♂️

src/components/SchemaRow.tsx
marcelltoth: The problem with that is exactly what you suggest. There is no 1-1 relationship between JST depth, and JSV depth because of array flattening. I don't think this prop adds much noise though, I think it's much more straightforward logic than trying to deduce the real depth from `schemaNode.depth` and its ancestor tree.

src/__fixtures__/arrays/of-objects.json
marcelltoth: Probably, from times when I was trying to make tests pass here. (Before changing direction and coming up with #111 ) 

Nice catch, thank you. Undone.

marcelltoth:
Oh, these magic words always get me, even though I know them 😄 Thanks for the heads-up @domagojk, I'll update it.

src/components/JsonSchemaViewer.tsx
```diff
@@ -1,127 +1,77 @@
...
+  onGoToRef,
+  renderRowAddon,
+}) => {
+  const jsonSchemaTreeRoot = React.useMemo(() => {
+    const jsonSchemaTree = new JsonSchemaTree(schema, {
+      mergeAllOf: true,
+      refResolver: resolveRef,
     });
-  }
+    jsonSchemaTree.walker.hookInto('filter', node => {
```
P0lip, line 33: You lost `stepIn` in the transition.
This is quite important, otherwise _really_ large schemas with a ton of $refs will be much slower to process than needed.
`stepIn` is needed, so that JST tree is not built upfront for the whole schema, but for actual parts that are currently rendered.
This is also one of the more notable factors contributing to the entire complexity of the former solution.
While, getting rid of virtualization is fine by me, we certainly should at least try to preserve that optimization.
By adding it back, we should be more or less on par with current v3.

P0lip (commented):
Looks good with the exception of `stepIn` being gone in the new code.
We should try to add it back. LMK if you have any concerns
Apologies I didn't spot this during the first pass.

P0lip:
Hmm, or maybe not.
This is what happens when I try to open one of our reference models.
image
image

Chromium is no different. :cry: 
image

Note - I modified the story to load only a single instance of JSV instead of two.

```ts
 .add('stress-test schema', () => (
    <>
      <div style={{ height: '100vh', overflowY: 'scroll' }}>
        <JsonSchemaViewer
          schema={stressSchema as JSONSchema4}
          defaultExpandedDepth={number('defaultExpandedDepth', 2)}
          onGoToRef={action('onGoToRef')}
        />
      </div>
    </>
  ))
```
It's likely to be something off with JST meeting some very recursive data, as the model I test it against has plenty of such.
I'll look into it now.

---

I got the fix for that JST issue, but even with that, it's still quite slow.
With profiling on and no CPU throttling it takes over a second for render, which is probably too much.
With CPU throttling set to 4x, it's over 7s.
Do note that I still test it having the above storybook tweak applied.

Now, in the case of current `origin/beta` tip the results are ~250ms and ~750ms respectively for CPU 4x throttling.

package.json
```diff
@@ -37,7 +37,6 @@
   "peerDependencies": {
     "@stoplight/markdown-viewer": "^4",
     "@stoplight/ui-kit": "^3",
```
P0lip, line 39: Do we still rely on UI-Kit in JSV?

P0lip (approved):
Aside from that perf issue, the code is good :+1:

package.json
marcelltoth: Unfortunately it's still a peer of MDV :\

marcelltoth: I removed the include from the storybook `_styles.scss`. That at least make Storybook rebuilds faster.... which is a win I guess... 😅

PR #110, opened 2021-03-04
feat: oneOf/anyOf as a dropdown, attempt no 2

marcelltoth:
Implements most of stoplightio/elements#667

The only missing piece is figuring out titles based on $ref names (see the issue), that will be handled separately. That will need to be a JST feature, as we simply don't have the requested info right now (beyond looking at `fragment`).

@P0lip Do you think I can simply default the `title` property (meaning it will be used *everywhere*), or should I keep it separately in another property and only use it here?


src/utils/printName.ts
```diff
@@ -41,14 +41,14 @@ function printArrayName(schemaNode: RegularNode): string | null {
...
+    return val !== null && val.length > 0 ? `array of ${val.join('s/')}s` : null;
   }
 
   if (isComplexArray(schemaNode)) {
     const firstChild = schemaNode.children[0];
-    return firstChild.title !== null
-      ? `${firstChild.title}[]`
-      : `${SchemaNodeKind.Array}[${firstChild.primaryType ?? firstChild.combiners?.join(',')}]`;
+    return firstChild.title
+      ? `array of ${firstChild.title}-s`
```
P0lip, line 50: Did you consider using https://github.com/plurals/pluralize?
Although I'm usually not a huge fan of adding libraries, this one might turn out to be useful.
Adding `s` by default won't be valid in all cases, i.e.
`watch` -> `watches`, `witch` -> `witches`, or `water`, `rice`, etc.

src/tree/utils.ts
```diff
@@ -1,9 +1,23 @@
...
 import { ComplexArrayNode, FlattenableNode, PrimitiveArrayNode } from './types';
 
-export const isParentNode = (node: SchemaNode) => isRegularNode(node) && !!node.children && node.children.length > 0;
+export type ChildNode = RegularNode | ReferenceNode | MirroredSchemaNode;
+
+export const isParentNode = (
+  node: SchemaNode,
+): node is RegularNode & {
+  children: ChildNode[] & { 0: ChildNode };
+} => isRegularNode(node) && !!node.children && node.children.length > 0;
```
P0lip, line 20: I suggest to create a separate function. This is tad misleading now.
A parent node may have no child, i.e.

```ts
"type": "object",
"properties": {}
```

src/components/SchemaRow/SchemaRow.tsx
```diff
@@ -80,15 +84,49 @@ export const SchemaRow: React.FunctionComponent<SchemaRowProps> = ({ schemaNode,
...
+                  <Select
+                    size="sm"
+                    options={choices.map((choice, index) => ({
+                      value: index,
+                      label: choice.title,
+                    }))}
+                    value={
+                      choices
+                        .indexOf(selectedChoice)
+                        .toString() /* toString to work around https://github.com/stoplightio/mosaic/issues/162 */
```
P0lip, line 125: hmmm. I think Select should not accept numbers in general, because the 'native' select accepts only string values if I am not mistaken.

src/components/SchemaRow/SchemaRow.tsx
```diff
@@ -80,15 +84,49 @@ export const SchemaRow: React.FunctionComponent<SchemaRowProps> = ({ schemaNode,
...
+                ) : null}
+
+                {schemaNode.subpath.length > 1 && schemaNode.subpath[0] === 'patternProperties' ? (
+                  <div className="sl-ml-2 sl-truncate sl-text-muted">(pattern property)</div>
+                ) : null}
+                {choices.length > 1 && (
+                  <Select
+                    size="sm"
+                    options={choices.map((choice, index) => ({
+                      value: index,
```
P0lip, line 119: So perhaps let's just do `String(index)`
```suggestion
                      value: String(index), // or toString()
```

src/components/SchemaRow/useChoices.ts
```diff
@@ -0,0 +1,87 @@
...
+import { isComplexArray, isParentNode } from '../../tree';
+import { printName } from '../../utils';
+
+type Choice = {
+  title: string;
+  type: SchemaNode;
+};
+
+function calculateChoiceTitle(node: SchemaNode, isPlural: boolean): string {
+  const complexObjectSuffix = isPlural ? '-s' : '';
```
P0lip, line 14: Ditto re usage of `pluralize`.

src/components/SchemaRow/useChoices.ts
```diff
@@ -0,0 +1,87 @@
...
+  const primitiveSuffix = isPlural ? 's' : '';
+  if (isRegularNode(node)) {
+    const realName = printName(node);
+    if (realName) {
+      return realName + complexObjectSuffix;
+    }
+    return (node.primaryType ?? 'object') + primitiveSuffix;
```
P0lip, line 21: I'm not sure you want to default. If there's no associated `primaryType` you probably want to display `any` instead of `object`.
```suggestion
    return (node.primaryType ?? 'any') + primitiveSuffix;
```

src/components/SchemaRow/useChoices.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+  if (isReferenceNode(node)) {
+    const lastPiece = last(node.value?.split('/') ?? []);
+    if (lastPiece) {
+      return lastPiece.split('.')[0] + complexObjectSuffix;
+    }
+    return '$ref' + primitiveSuffix;
+  }
+
+  return 'object' + primitiveSuffix;
```
P0lip, line 31: Ditto. Defaulting to `object` might be potentially incorrect.

src/components/SchemaRow/useChoices.ts
```diff
@@ -0,0 +1,87 @@
...
+  const primitiveSuffix = isPlural ? 's' : '';
+  if (isRegularNode(node)) {
+    const realName = printName(node);
+    if (realName) {
+      return realName + complexObjectSuffix;
+    }
+    return (node.primaryType ?? 'object') + primitiveSuffix;
+  }
+  if (isReferenceNode(node)) {
+    const lastPiece = last(node.value?.split('/') ?? []);
```
P0lip, line 24: You probably want to use `pointerToPath` to handle `~0` and ~1`.

P0lip:
> @P0lip Do you think I can simply default the title property (meaning it will be used everywhere), or should I keep it separately in another property and only use it here?

Do you mean the potential JST change?

Based on the following sentence

> The only missing piece is figuring out titles based on $ref names (see the issue), that will be handled separately

and the actual requirement the above adhered to

>  the last segment of the $ref. If last segment is a file, chop off the extension

I'm not sure we need any change.

What you have implemented here seems to be covering everything.
Am I missing anything?

src/components/SchemaRow/useChoices.ts
P0lip: It's in `@stoplight/json`.

src/utils/printName.ts
marcelltoth: I did consider that exact library, but I realized I don't need it here:
- If it's an object name (like in this case), I much prefer to have the hyphen, because it emphasizes that those are entities with a name. Let's say I have an object titled `Integer` (for some effed up reason). If it says *array of Integer-s*, it's clear that it's different from e.g. *array of numbers*.
- For primitives, fortunately a simple `s` fits them all.
- For objects, I really don't want the English rules. Think about an object called *Person*. I don't want the array to say *array of People* 😄 I agree re *array of Watch-s* isn't ideal but it gets the point across, and that case would be hard to cover even with the library (it doesn't do hyphenation). I'd say that's an improvement for later.

src/tree/utils.ts
marcelltoth: Will do

src/components/SchemaRow/SchemaRow.tsx
marcelltoth: That is a solution, too, although I much prefer `react-select`s signature, which is generic, then it's much easier to work with:
`<Select items={choices} onChange={setSelectedChoice} />` 

Accessibility suffers though, that's true. Either way is fine for me, but the types should match the behavior.

src/components/SchemaRow/useChoices.ts
marcelltoth: Let me close this and discuss above.

src/components/SchemaRow/useChoices.ts
marcelltoth: Good point, I'll correct.

philsturgeon:
Can you give a before and after screenshot so I can approve? Thanks!

src/components/SchemaRow/useChoices.ts
marcelltoth: Thank you, I'll do that

marcelltoth:
@P0lip The scenario that you're missing:

Take the following schema:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": {
          "type": "string"
        }
      },
      "required": [
        "street_address"
      ]
    },
    "person": {
      "type": "object",
      "properties": {
        "first_name": {
          "type": "string"
        },
        "last_name": {
          "type": "string"
        }
      }
    }
  },
  "type": "object",
  "properties": {
    "combiner": {
      "oneOf": [{"$ref": "#/definitions/address"}, {"$ref": "#/definitions/person"}]
    },
    "billing_address": {
      "$ref": "#/definitions/address"
    },
    "shipping_address": {
      "$ref": "#/definitions/address"
    }
  }
}
```

Now if I render this, this currently lets me chose between "object" and well... "object".
image


What @philsturgeon was looking for in the issue description is that I realize that these objects are referenced as `#/definitions/person` and automatically name it "person". 

My question to you both is: 

**Should this be a feature for this specific dropdown, or can it be a general feature?** 

To give you an example based on the above, if this can be made a generic JSV feature (which is the simpler one to implement actually I believe), then the above schema would render like this:
image

philsturgeon:
If we cannot handle finding names when there's no title we can just call them `object #1` and `object #2`.

marcelltoth:
> If we cannot handle finding names when there's no title we can just call them `object #1` and `object #2`.

That's OK, but that doesn't answer the question if we can - in general - use the `$ref` names for objects that lack a `title`?

philsturgeon:
We are not trying to remove object and replace it with title everywhere, we are trying to offer a drop down that makes sense to people.

src/tree/utils.ts
```diff
@@ -1,9 +1,23 @@
...
 import { isNonNullable } from '../guards/isNonNullable';
 import { ComplexArrayNode, FlattenableNode, PrimitiveArrayNode } from './types';
 
-export const isParentNode = (node: SchemaNode) => isRegularNode(node) && !!node.children && node.children.length > 0;
+export type ChildNode = RegularNode | ReferenceNode | MirroredSchemaNode;
+
+export const isNonEmptyParentNode = (
+  node: SchemaNode,
+): node is RegularNode & {
+  children: ChildNode[] & { 0: ChildNode };
```
mpodlasin, line 19: lol. This blew my mind. :joy:

P0lip: Yeah, unfortunately TS does not have any other way to accomplish this so we end up using that in a few projects :laughing:

src/components/SchemaRow/useChoices.ts
```diff
@@ -0,0 +1,89 @@
...
+  const primitiveSuffix = isPlural ? 's' : '';
+  if (isRegularNode(node)) {
+    const realName = printName(node);
+    if (realName) {
+      return realName + complexObjectSuffix;
+    }
+    return node.primaryType !== null ? node.primaryType + primitiveSuffix : 'any';
+  }
+  if (isReferenceNode(node)) {
+    const lastPiece = last(pointerToPath(node.value ?? ''));
```
P0lip, line 25:
```suggestion
    const lastPiece = !node.error && node.value ? last(pointerToPath(node.value)) : null;
```
`pointerToPath` will throw an exception if invalid $ref is given.

P0lip (approved):
Looks good apart from this one issue.

I didn't verify the UI or functional aspects of this PR, though.
Someone from undefined team should do that

mallachari:
I was trying to break it with multiple weird nested schemas and it behaves well. Only thing that brought my attention was an array of anyOf/oneOf:

Screenshot from 2021-03-24 13-49-26

Do we have defined how to render such cases? I guess it would require multiple select inputs

src/tree/utils.ts
marcelltoth: > *praying no one mentions the word Monad 🙏*

marcelltoth:
> 
> 
> I was trying to break it with multiple weird nested schemas and it behaves well. Only thing that brought my attention was an array of anyOf/oneOf:
> 
> Screenshot from 2021-03-24 13-49-26
> 
> Do we have defined how to render such cases? I guess it would require multiple select inputs

This is weird, the simple use-case should work. What is the exact schema you are testing with? I've tried changing the example schema I sent above to 
```json

  "properties": {
    "combiner": {
      "type": "array",
      "items": {
        "oneOf": [{"$ref": "#/definitions/address"}, {"$ref": "#/definitions/person"}]
      }
    },
    ...
```

And it seems to work just fine:
image

This doesn't require nested selects though. The more complex case that does is handled separately at https://github.com/stoplightio/elements/issues/713

mallachari:
> This is weird, the simple use-case should work. What is the exact schema you are testing with? 

I used this type of schema:
```json
{
  "type": "array",
  "items": {
    "oneOf": [
      {
        "type": "object",
        "properties": {
          "aaa": {
            "type": "string"
          }
        }
      },
      {
        "type": "object",
        "properties": {
          "bbb": {
            "type": "string"
          }
        }
      }
    ]
  }
}
```

Though it is nested in another anyOf combiner so exactly the case that you mentioned of multiple depth anyOf/oneOf. It works the way you described if it's the plain property. 
So it's not in scope of this issue in the end.

PR #113, opened 2021-03-18
perf: improve performance

marcelltoth:
Times in parenthesis are mounting times of the 2x-stress schema story in Storybook.
- Bumps @stoplightio/json-schema-merge-allof (1660ms -> 1590ms)
- Removes trivial mosaic components from the performance-critical rendering paths: `SchemaRow` and beyond (1590ms -> 850ms)

The problem with mosaic components is that for almost 0 gain (beyond type-checking of style-props), you render yet another element, and especially yet another `forwardRef`-ed element, both of which eat up performance as you scale.

### DIFF
--- a/src/__stories__/JsonSchemaViewer.tsx
+++ b/src/__stories__/JsonSchemaViewer.tsx
@@ -63,22 +63,11 @@ storiesOf('JsonSchemaViewer', module)
     );
   })
   .add('stress-test schema', () => (
-    <>
-      <div style={{ height: 345, overflowY: 'scroll' }}>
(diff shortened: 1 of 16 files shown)


P0lip (approved):
:+1: perhaps someone from undefined-team should sign off the UI part of this PR.
It looks good to me, yet I shouldn't be the one deciding whether you're dropping certain mosaic components or not. However as an external person, I'd say it's a reasonable choice.

mpodlasin (approved):
I wasn't able to detect any visual regressions, so I believe this is ok!

PR #117, opened 2021-03-26
fix: styling issues

marcelltoth:
1. Symmetrical spacing on left and right sides
2. Descriptions no longer have `white-space: nowrap` which would introduce horizontal scrollbars.
3. Top 2 carets actually line up
image


P0lip (approved):
:+1:

P0lip:
Do not forget to update snapshots.

marcelltoth:
Ah, I never think about updating tests for stylistic changes.

PR #118, opened 2021-03-29
feat: prefix choices with an index

marcelltoth:
Last tiny fix before JSV goes live in Elements. 

This is to mitigate that we don't retain original $ref-s in nodes (which I will implement in JST soon, but still this is a short fix).


mpodlasin (approved):

P0lip (approved):
:shipit:

PR #121, opened 2021-04-01
fix: bump json-schema-tree to 1.1.2

mmiask:
A part of fixing: https://github.com/stoplightio/elements/issues/989 - Elements hang with the new JSV on board, if we open a complex model containing multiple `$refs` and `allOf`s

This bumps the `json-schema-tree` that recently bumped `json-schema-merge-allof` which is fixing that problem (we checked by manually modifying the lockfile to use the latest `json-schema-merge-allof`)


P0lip (approved):
:+1:

marcelltoth:
@P0lip we are investigating whether the included snapshot change was intentional. Can you double check please?

P0lip:
I checked it yesterday when I was releasing JST@1.1.2, and it's good.
I did side-by-side comparison and didn't notice any UI change.
The reason the snapshots are different is because the $ref handling has improved, and as a result we have more mirror nodes.
Since JSV does not expand mirrored nodes by default, the snapshots have been updated, since we only dump the initial state of component, without any user interaction.
From user perspective, nothing should change, perhaps with the exception that some node might be now collapsed, and they weren't before.
This is not a new behavior, however, as $refs were always supposed to be collapsed by default.

marcelltoth:
Amazing, thank you!

PR #126, opened 2021-04-07
chore: export validations

mallachari:
Exports Validations component.
Changes display of boolean validations and puts them in one row instead of separate divs.

### DIFF
--- a/src/components/shared/Validations.tsx
+++ b/src/components/shared/Validations.tsx
@@ -137,18 +137,22 @@ const KeyValueValidation = ({ name, values }: { name: string; values: string[] }
 
 const NameValidations = ({ validations }: { validations: Dictionary<unknown> }) => (
   <>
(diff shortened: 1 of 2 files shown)


marcelltoth (commented):
@mallachari First of all, if you would like to trigger a release you should pick either the `feat` or `fix` keywords, this would probably be a `feat`.

OTOH let's not forget that JSV is a public library, so exposing components is a big deal. If we decide to pursue this route, we should proceed with caution.

P0lip:
Hmm, couldn't a nested import be used?
I second Marcell here, and I'd also a bit skeptical about exposing such components publicly.

mallachari:
Nested one didn't work for me when I tried that but I'll check again. Anyway I'm closing that pr as there's another one open and if needed it can be added there.

Issue #119, opened 2021-03-29
Batch of JSV UI/UX Fixes


marcelltoth:
## Describe the bug

**1. Border for parameters runs too far to the left**

Right now if I embed JSV in Elements, I get an ugly left padding, see Screenshot. In stoplightio/elements#979 I am forced to use some negative margin to work around.

The component should be exactly as large as the space it actually uses, just like any well-behaving element.

Currently looks like this:

![image](https://user-images.githubusercontent.com/543372/112846577-7eb9aa80-90a6-11eb-93bb-abfb1ee9f8f3.png)

Should look like this:

<img width="446" alt="image" src="https://user-images.githubusercontent.com/67381/113162861-8c526a00-9237-11eb-9e85-3f1771055ec6.png">

**2. There should not be more space between title and description than there is padding at top and bottom of a row**

Currently:

<img width="478" alt="image" src="https://user-images.githubusercontent.com/67381/113164559-1e0ea700-9239-11eb-9c46-e9e6db3948a0.png">

Spacing could be closer to the Figma:

<img width="641" alt="image" src="https://user-images.githubusercontent.com/67381/113163714-519d0180-9238-11eb-8f12-2d1efa4eed94.png">

**3. Background color for examples / allowed values should be `#EDF2F7`**

Currently 

![image](https://user-images.githubusercontent.com/67381/113166616-d7ba4780-923a-11eb-9b18-dfe9ac38ddce.png)

Should be 

<img width="454" alt="image" src="https://user-images.githubusercontent.com/67381/113166823-0c2e0380-923b-11eb-9cfb-568de77c1546.png">

PR #127, opened 2021-04-07
fix: ui fixes

mallachari:
Fixes https://github.com/stoplightio/json-schema-viewer/issues/119

Removed left margin for top rows so it spans to the edge. That still might require small negative margin in elements to compensate the caret.

Screenshot from 2021-04-07 12-31-00


mpodlasin (changes requested):
Arrow and text are misaligned (arrow should be a bit lower here)

Screenshot from 2021-04-07 16-25-53

Property name and type are misaligned (type should be a bit higher)

Screenshot from 2021-04-07 16-25-43

mallachari:
> Arrow and text are misaligned (arrow should be a bit lower here)
> 
> Screenshot from 2021-04-07 16-25-53
> 
> Property name and type are misaligned (type should be a bit higher)
> 
> Screenshot from 2021-04-07 16-25-43

That arrow thing is included in another issue, not usre about second one, but those should be fairly simple so I can fix it here.

mpodlasin (approved):

mallachari:
> Ah sorry, I thought that this is the last UI in JSV issue. I am accepting then. Please attempt to fix the name/type issue, but if it's too hard or something, create an issue for it or just add it to that last "UI in JSV" we have.

Fixed both :)

PR #128, opened 2021-04-07
fix: handling nested schemas with allof

mpodlasin:
Relates to https://github.com/stoplightio/elements/issues/1006


P0lip (approved):
:+1:

PR #129, opened 2021-04-07
fix: use items validation for flattened arrays

P0lip:
Addresses https://github.com/stoplightio/platform-internal/issues/5799

image
I'm so happy this code is entirely gone in beta, and that JST has it done more reasonably :see_no_evil: 

`tickets.schema.json` has been updated due to that fix, see the portion of that whole schema document
```json
    "ticketRecipients": {
                            "type": "array",
                            "items": {
                              "type": "string",
                              "enum": [
                                "BOOKER",
                                "CUSTOMER",
                                "PASSENGER",
                                "THIRD_PARTY"
                              ]
                            }
                          },
```

### DIFF
--- a/src/tree/utils/walk.ts
+++ b/src/tree/utils/walk.ts
@@ -49,10 +49,24 @@ export function* processNode(node: JSONSchema4): IterableIterator<SchemaNode> {
       yield combinerNode;
     }
   } else if (type) {
+    const primaryType = getPrimaryType(node);
+    let validationNOde: JSONSchema4 = node;
+    if (
+      primaryType === SchemaKind.Array &&
+      _isObject(node.items) &&
+      !Array.isArray(node.items) &&
+      getCombiners(node.items) === void 0
+    ) {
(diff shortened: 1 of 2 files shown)


src/tree/utils/walk.ts
```diff
@@ -49,10 +49,24 @@ export function* processNode(node: JSONSchema4): IterableIterator<SchemaNode> {
       yield combinerNode;
     }
   } else if (type) {
+    const primaryType = getPrimaryType(node);
+    let validationNOde: JSONSchema4 = node;
```
paulatulis, line 53: is there a reason this is `validationNOde` and not `validationNode`?

paulatulis (commented):
Looks good! I pulled down the branch and tested with the example you show in the summary (thanks for including), and it does what it says it should. I just have one question about the variable name.

src/tree/utils/walk.ts
P0lip: Ah yeah. Typo.

paulatulis (approved):
Bueno

PR #130, opened 2021-04-07
feat: do not display min/max values for oas formats

P0lip:
Addresses https://github.com/stoplightio/platform-internal/issues/4350

Initially I wanted to address this directly in `http-spec`, but I quickly changed my mind, as this could potentially break certain usages. These 5 formats are OAS-specific, and JSON Schema does not support them, therefore the conversion we do includes correct ranges, so that generic JSON Schema tooling can interpret such a schema correctly.
Unfortunately, as a side effect, the values get inserted to the document, and JSV renders them.
The solution I decided to go with is rather simple. Since a format implies the valid ranges, we do not need to show it unless a given value was overridden manually.

To be ported to JST / JSV v4.

### DIFF
--- a/src/components/__tests__/Validations.spec.tsx
+++ b/src/components/__tests__/Validations.spec.tsx
@@ -32,6 +32,8 @@ describe('Validations component', () => {
       expect(wrapper).toHaveText('deprecated');
     });
   });
+
+  test.todo('when oas format is specified');
 });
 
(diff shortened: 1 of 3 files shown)


paulatulis (approved):
This works for me, so approving. If anyone else on @stoplightio/void-crew wants to take a look to sanity check me, please do!

PR #131, opened 2021-04-08
feat: do not display min/max values for oas formats

P0lip:
Cherry-pick of https://github.com/stoplightio/json-schema-viewer/pull/130


wmhilton (approved):
LOLOL I thought that might be the approach. Insert crap min/max the beginning of the pipeline, filter out crap min/max at the end of the pipeline... :trollface:

Issue #122, opened 2021-04-01
Mega Batch of JSV UI/UX fixes


philsturgeon:
Another few items from the UI/UX list, items 16 and 17 are labelled 16 and 17 in the image below.

16. Border color is swapped - operation title bottom border (for request, response, etc) should be default border color and JSV divider color should be sl-border-light (the more prominent UI element on the page should have a higher contrast border)

17. JSV: expand/collapse caret is slightly mis-aligned with the text in it's row

![image.png](https://images.zenhubusercontent.com/5d1f7c5b8a1fa426d066830f/75cd793c-0a77-4a72-b113-c4967c9f5a0d)

These bits are not labelled in the image:

- On schema pages, do not render the root level schema description in the JSV, since we're already rendering it above the schema
- On schema page, remove the border around the JSV. We can remove the ability to show examples.
- Property descriptions should be at base font size (same as property title)

PR #132, opened 2021-04-08
fix: divider

mallachari:
As part of https://github.com/stoplightio/json-schema-viewer/issues/122 (no 16) we need to have divider in light color. 
Yes, it could go in in previous pr but that was another issue and I didn't see it ;)


marcelltoth (approved):
Save the world @mallachari , save it! 🚀 😄

PR #134, opened 2021-04-14
feat: hide examples option

mallachari:
Fixes https://github.com/stoplightio/elements/issues/1026

Adds `hideExamples` property to JSV that prevents rendering example validations.

### DIFF
--- a/src/__stories__/JsonSchemaViewer.tsx
+++ b/src/__stories__/JsonSchemaViewer.tsx
@@ -1,6 +1,6 @@
 import { Button, Flex, InvertTheme, subscribeTheme } from '@stoplight/mosaic';
 import { action } from '@storybook/addon-actions';
-import { number, object, select, withKnobs } from '@storybook/addon-knobs';
+import { boolean, number, object, select, withKnobs } from '@storybook/addon-knobs';
 import { storiesOf } from '@storybook/react';
 import { JSONSchema4 } from 'json-schema';
 import * as React from 'react';
@@ -35,6 +35,7 @@ storiesOf('JsonSchemaViewer', module)
         },
         'standalone',
       )}
+      hideExamples={boolean('hideExamples', false)}
       onGoToRef={action('onGoToRef')}
     />
   ))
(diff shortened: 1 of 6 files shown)


mallachari:
@P0lip adding that property was an issue requirement, is it acceptable from your side? We were considering just modifying schema object but that would require lot of additional processing.

P0lip:
Yeah, seems okay.

>  We were considering just modifying schema object but that would require lot of additional processing.

Hmm, it should be quite easy actually.

```ts
  jsonSchemaTree.walker.on('exitNode', node => {
      if (!isRegularNode(node)) return;
      for (const field of ['examples', 'example', 'x-example']) {
         if (Array.isArray(node.annotations[field])) {
           node.validations[field] = node.annotations[field];
           delete node.annotations[field];
        }
    });
```
`annotations` and `validations` might be readonly though, so um, TS will scream :D

paulatulis (approved):
When I pulled this down and played around with it, it worked for me, so I am going to approve, but @P0lip it might make everyone feel more secure if you also give your stamp of approval ;) 

I guess my only more general question is related to the issue and understanding why this particular thing would necessitate an extra prop when other items are deemed appropriate to alter the schema. Maybe this doesn't come up very often and it looks like it was already sussed out in the issue anyway. Just curious if this is a common occurrence.

mallachari:
> I guess my only more general question is related to the issue and understanding why this particular thing would necessitate an extra prop when other items are deemed appropriate to alter the schema. Maybe this doesn't come up very often and it looks like it was already sussed out in the issue anyway. Just curious if this is a common occurrence.

The reason is that we need to have possibility to not display example validations from Elements to not duplicate it as it already includes examples in different component. We could do it by feeding jsv with modified schema but that would need to be done from elements and require some additional computation on each schema change that might decrease performance. That is the only optional validation that might exist so we definitely won't be adding any more weird props.

paulatulis:
> The reason is that we need to have possibility to not display example validations from Elements to not duplicate it as it already includes examples in different component. We could do it by feeding jsv with modified schema but that would need to be done from elements and require some additional computation on each schema change that might decrease performance. That is the only optional validation that might exist so we definitely won't be adding any more weird props.

That makes sense! Thanks @mallachari :)

PR #135, opened 2021-04-14
feat: special treatment for top level rows

marcelltoth:
Implements stoplightio/elements#1030


mpodlasin (commented):
LGTM

mpodlasin:
> Didn't have time to check root level use cases beyond the one in demo w array - @philsturgeon over to you to try to verify as many root situations as possible (that they are handled well and are easy to understand as end user).
> 
> @marcelltoth two small UI things:
> 
>     1. root prop name should be same size as other props (it's 11px versus 12px - in general i'm not sure if setting things to text size small so high up in tree, and then overriding to base is best since most of it should be base size, but up to you).
> 
>     2. font weight for prop names should be semibold, not bold
> 
> 
> Screen Shot 2021-04-15 at 11 26 49 AM

1. I changed font sizes
2. Changed bold to semibold, but in this font the change is not visible, next available weight is medium, but it looks too thin.

marbemac:
> Changed bold to semibold, but in this font the change is not visible, next available weight is medium, but it looks too thin.

That's OK - please leave as semibold, we'll be using Inter in our applications and it is a free font so will document it as the recommended font to pair Elements with.

src/components/SchemaRow/TypeSelector.tsx
```diff
@@ -0,0 +1,24 @@
+import { Select } from '@stoplight/mosaic';
+import * as React from 'react';
+
+import type { useChoices } from './useChoices';
+
+type TypeSelectorProps = ReturnType<typeof useChoices>;
+
+export const TypeSelector = ({ choices, setSelectedChoice, selectedChoice }: TypeSelectorProps) => {
```
mallachari, line 8: What's the point of this component? It is not used anywhere here from what I can see.

mallachari (approved):
Looks better and works well. I couldn't break it or find any edge case.

src/components/SchemaRow/TypeSelector.tsx
mpodlasin: @marcelltoth some info on this? @mallachari is right, it doesn't seem to be used at all.

marcelltoth: Oh, right. That was an attempt to extract the common selector from `SchemaRow` and `TopLevelSchemaRow` but it turned out that the two are different so it didn't make sense to do it in the end. I removed it now, thanks @mallachari for spotting it.

PR #137, opened 2021-04-23
fix: pin mosaic to exact version

mmiask:
This is connected to https://github.com/stoplightio/elements/issues/1051 - the issue turned out to not happen on `main` once we did a fresh install, but it made us think that it might be a good idea to pin Mosaic to specific version before its stable version is released.


marcelltoth (approved):
You convinced me @mmiask @mpodlasin

PR #138, opened 2021-04-26
feat: accept JSON Schema Draft 6 & Draft 7

P0lip:
For https://github.com/stoplightio/platform-internal/pull/6083
This _does not_ add any meaningful support of new features introduced in Draft 6 & Draft 7. Everything's going to be processed in the same way as it used to be.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,12 +42,14 @@
     "react-dom": ">=16.8"
   },
   "dependencies": {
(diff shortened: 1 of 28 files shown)


wmhilton (approved):
Doesn't _look_ like anything could go wrong :trollface: I like that the typings resulted in some more edge case handling around booleans lol.

PR #139, opened 2021-04-26
feat: accept JSON Schema Draft 6 & Draft 7

P0lip:
Needed by https://github.com/stoplightio/platform-internal/pull/6083
Needs https://github.com/stoplightio/json-schema-tree/pull/8
For https://github.com/stoplightio/elements/issues/1116

See my comments to learn more about the changes.

### DIFF
--- a/src/__tests__/__snapshots__/index.spec.tsx.snap
+++ b/src/__tests__/__snapshots__/index.spec.tsx.snap
@@ -1068,13 +1068,13 @@ exports[`HTML Output should match arrays/of-complex-objects.json 1`] = `
                   </div>
                 </div>
                 <div>
+                  <span>Allowed value:</span>
+                  <span style=\\"background-color: rgb(237, 242, 247)\\">\\"Constant name\\"</span>
+                </div>
+                <div>
                   <span>Example values:</span>
                   <span style=\\"background-color: rgb(237, 242, 247)\\">\\"Example name\\"</span>
                   <span style=\\"background-color: rgb(237, 242, 247)\\">\\"Different name\\"</span>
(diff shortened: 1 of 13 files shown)


src/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -1067,15 +1067,15 @@ exports[`HTML Output should match arrays/of-complex-objects.json 1`] = `
                     </div>
                   </div>
                 </div>
+                <div>
```
P0lip, line 1070: I hope the order doesn't matter for you.

src/components/shared/Validations.tsx
```diff
@@ -25,7 +25,7 @@ export const numberValidationNames = [
   'exclusiveMaximum',
 ];
 
-const exampleValidationNames = ['examples', 'example', 'x-example'];
+const exampleValidationNames = ['examples'];
```
P0lip, line 28: `http-spec` will try to produce `examples` at all times now, thus there's no need to handle `example` and `x-example` separately.

src/components/shared/Validations.tsx
```diff
@@ -58,11 +58,8 @@ const createValidationsFormatter = (name: string, options?: { exact?: boolean; n
 };
 
 const validationFormatters: Record<string, (value: unknown) => ValidationFormat | null> = {
-  ['const']: createValidationsFormatter('Allowed'),
```
P0lip, line 61: `json-schema-tree` turns `const` into an enum with a single member

domagojk (approved):
LGTM

PR #147, opened 2021-05-31
fix: dark mode validations

mallachari:
Needed for: https://github.com/stoplightio/elements/issues/1149

Makes validations values support dark mode

Screenshot from 2021-05-31 13-06-51


mpodlasin (commented):
I know that @marcelltoth was removing some mosaic elements from JSV to reduce render times.

Maybe you two can collaborate to test if/how this change affects performance?

Unless @marcelltoth believes that it will be a non-issue here of course :)

marcelltoth:
> I know that @marcelltoth was removing some mosaic elements from JSV to reduce render times.
> 
> Maybe you two can collaborate to test if/how this change affects performance?

That was my only worrying thought as well. 😄 Let's see how the stress schema renders. Maybe it's not that bad as it's only validations. 🤔

mallachari:
> > I know that @marcelltoth was removing some mosaic elements from JSV to reduce render times.
> > Maybe you two can collaborate to test if/how this change affects performance?
> 
> That was my only worrying thought as well.  Let's see how the stress schema renders. Maybe it's not that bad as it's only validations.

Well, it seems to render fine from what I can see. I added those mosaic components by the way, so we can keep it the way it was, just use it for validations where it's necessary.

marcelltoth (approved):
I did a rough measurement now. It's worse, but not horrible, around ~5-10%. I'm OK with it if you are.

PR #149, opened 2021-06-02
fix: remove ui-kit

mallachari:
Needed for https://github.com/stoplightio/elements/pull/1360

- Removes ui-kit dependency
- Uses new markdown-viewer
- **Uses new sl-scripts** - this is the only solution I found that made JSV work with markdown-viewer when used outside with webpack 4. More specifically create-react-app in one of elements examples was using jsv as es module which was then importing mdv from cjs module which was breaking build. Building JSV the same way as MDV with recent scripts makes it work perfectly.


wmhilton:
Oh this looks great! Let's yalc it into ninja/studio just to derisk it. I can do it later this evening if you want

wmhilton:
Eh, it doesn't quite look right yalced into platform, but I think it's just a margins and borders. 

image


image

I suppose I need to yalc https://github.com/stoplightio/elements/pull/1360 as well to get the full picture. Uno momento...

wmhilton:
Ah. I can't seem to yalc `elements-core` into platform bc it complains `Cannot find module '@stoplight/elements-core/containers/TableOfContents'`

wmhilton (approved):
Code looks fine, Storybook looks fine. Yalc looks off but the Storybook looks fine, so that's likely just an integration / usage issue in platform.

mallachari:
Thanks for checking. It possibly needs some simple styling in studio due to missing ui-kit classes and mdv using different font-size. Though it should come with elements once https://github.com/stoplightio/elements/pull/1332 is merged.

package.json
```diff
@@ -35,30 +35,28 @@
...
     "lodash": "^4.17.19"
   },
   "devDependencies": {
-    "@rollup/plugin-typescript": "^3.1.1",
     "@sambego/storybook-state": "^1.3.6",
     "@stoplight/eslint-config": "^1.2.0",
-    "@stoplight/markdown-viewer": "^4.3.4",
-    "@stoplight/scripts": "^8.2.0",
+    "@stoplight/markdown-viewer": "^5.0.0-beta.4",
+    "@stoplight/scripts": "9.0.1",
```
P0lip, line 57: Are you sure the package can be loaded in node.js-compliant env?
Could we add https://github.com/stoplightio/markdown/blob/beta/package.json#L40?

P0lip (changes requested):
Looks good in general, but let's make sure the package can actually be loaded.

package.json
mallachari: I've added it and looks it's breaking for lodash:
```
SyntaxError: Named export 'last' not found. The requested module 'lodash' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export
```
Though I can see it's also used in markdown so I'll see how it's done there

mallachari: Changing lodash imports to default ones: `import last from 'lodash/{function}.js';`, same way it's done in `markdown` made it work. I just had to add `allowSyntheticDefaultImports`. Would that be enough?

P0lip (approved):

PR #151, opened 2021-06-21
feat: better names for oneOf/anyOf selector

### LINKED ISSUE #125: Generate titles based on $ref names for oneOf/anyOf dropdown
** User Story**

As a docs viewer I like to see descriptive names for switching between oneOf/anyOf, even if the author has not used the `title` property in every schema.

**Details**

Currently when we do oneOf/anyOf dropdowns we only look for the `title` property, and if there is no title we will show "1. object" and "2. object". This issue provides some logic for generating titles from a $ref if its provided, by looking at the last segment.

The last segment of the $ref should usually be ok to use as a "display name", which will show up in that dropdown.

- `$ref: #/components/schemas/Something` = Something
- `$ref: #/components/schemas/SomethingElse` = SomethingElse
- `$ref: #/components/schemas/SomethingElse/properties/something_else` = something_else
- `$ref: shared/turtle.json` = turtle
- `$ref: #/components/headers/Sunset` = Sunset

So assuming none of those refs have a title property inside, if a schema contains this:

```
oneOf: 
- $ref: "shared/turtle.json"
- $ref: "#/components/headers/Sunset"
- type: number
```

Then the dropdown should show:

- turtle
- Sunset
- number

Titles should still be taken from the schema object when set:

```
// openapi.yaml

components: 
  schema:
    Tag:
      title: Tag Model
      type: object
      properties:
        id:
          # ...
```

These titles can exist in "shared schemas" too:

```
// turtle.json

{
  "title": "Custom Turtle Model Name",
  "type": "object"
}
```

So look for `title`, if there's no `title` then take the last $ref segment and display it unmodified in whatever case with hyphens or whatever else.

_Also currently we have "1. Foo" and "2. Bar" but it should not be a numbered list, just a select box. We only add numbers if there are unknown types like "object #1" and "object #2"._

**Additional Context**

#110

marcelltoth:
Resolves #125 

Depends on https://github.com/stoplightio/json-schema-tree/pull/11

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "dependencies": {
     "@fortawesome/free-solid-svg-icons": "^5.15.2",
     "@stoplight/json": "^3.10.0",
-    "@stoplight/json-schema-tree": "^2.0.1",
+    "@stoplight/json-schema-tree": "^2.1.0",
     "@stoplight/react-error-boundary": "^1.0.0",
     "@types/json-schema": "^7.0.7",
     "classnames": "^2.2.6",
--- a/src/__fixtures__/combiners/oneof-within-array-item.json
+++ b/src/__fixtures__/combiners/oneof-within-array-item.json
@@ -15,17 +15,21 @@
         }
       },
       {
-        "type": "object",
-        "title": "B",
-        "properties": {
-          "foo": {
-            "type": "number"
-          },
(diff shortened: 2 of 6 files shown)


mallachari (approved):
Good feature!

Issue #141, opened 2021-05-05
Promote JSV v4 to stable


marcelltoth:
As agreed, we can mark JSV v4 as stable, releasing 4.0.0 and tagging it _latest_.

marcelltoth:
Do we want to hold off with this until mosaic is stable? @mallachari @P0lip

P0lip:
If there are not breaking changes coming up in Mosaic that would affect JSV, I believe we can make JSV v4 stable.

mallachari:
Should be good to go. Nothing breaking that I can see

PR #153, opened 2021-06-22
feat: depend on mosaic 1.0.0

marcelltoth:
Nice to have for #141


package.json
```diff
@@ -39,7 +39,8 @@
   },
   "peerDependencies": {
     "@stoplight/markdown-viewer": "^5.0.0-beta.5",
-    "@stoplight/mosaic": "^1.0.0-beta.59",
+    "@stoplight/mosaic": "^1.0.0",
+    "@stoplight/mosaic-code-viewer": "^1.0.0",
```
marcelltoth, line 43: This was missing. Needed for (peer of) MDV.

mallachari (approved):

mallachari:
Hmm, what's wring with those types? Is it some react incompatibility?

paulatulis:
Should the test_and_release workflow pass or is that something that can be ignored?

Also, is there somewhere I could see this in action? If not, since it's a nice-to-have, I will defer to @mallachari's review since he is probably more familiar with what this affects.

marcelltoth:
> 
> 
> Should the test_and_release workflow pass or is that something that can be ignored?

It absolutely should, updated now!

> Also, is there somewhere I could see this in action? If not, since it's a nice-to-have, I will defer to @mallachari's review since he is probably more familiar with what this affects.

Unfortunately no. JSV doesn't have PR environments AFAIK so you have to run it locally. Also there should be no visible change compared to current `beta`.

PR #154, opened 2021-06-24
fix: choice naming

marcelltoth:
Resolves @philsturgeon 's issues mentioned at https://stoplight-internal.slack.com/archives/C01D80M6EHX/p1624548332462800


src/components/SchemaRow/useChoices.ts
```diff
@@ -12,12 +12,11 @@ type Choice = {
 };
 
 function calculateChoiceTitle(node: SchemaNode, isPlural: boolean): string {
-  const complexObjectSuffix = isPlural ? '-s' : '';
   const primitiveSuffix = isPlural ? 's' : '';
```
philsturgeon, line 15: Let's get rid of this too? puralization is hard in English and putting an s on the end of things doesn't always work.

Oppertunitys
Monkeys

Doesn't work. :)

marcelltoth: That second one does work AFAICT. It's for primitives, I tried them all, they all luckily work. *objects*, *integers*, *booleans*, *nulls*, *strings*.

I believe it feels much more natural to have that pluralization there, but LMK if you can reproduce a broken one.

philsturgeon:
Ah ok cool. Yeah that’s fine. 

--
Phil Sturgeon
Product @ Stoplight.io

P0lip (approved):
The code looks :ok_hand:

PR #162, opened 2021-08-13
feat: do not wrap string validations

mallachari:
Addresses: https://github.com/stoplightio/platform-internal/issues/7377

Quotes were added to distinguish the type of validation. 
This PR makes `enum`, `examples` and `default` not show quotes regardless of type.

### DIFF
--- a/src/components/shared/Validations.tsx
+++ b/src/components/shared/Validations.tsx
@@ -63,11 +63,11 @@ const createValidationsFormatter = (name: string, options?: { exact?: boolean; n
 };
 
 const validationFormatters: Record<string, (value: unknown) => ValidationFormat | null> = {
-  enum: createValidationsFormatter('Allowed'),
-  examples: createValidationsFormatter('Example'),
+  enum: createValidationsFormatter('Allowed', { nowrap: true }),
+  examples: createValidationsFormatter('Example', { nowrap: true }),
   multipleOf: createValidationsFormatter('Multiple of', { exact: true }),
   pattern: createValidationsFormatter('Match pattern', { exact: true, nowrap: true }),
-  default: createValidationsFormatter('Default'),
+  default: createValidationsFormatter('Default', { nowrap: true }),
 };
 
 const oasFormats = {
--- a/src/__tests__/__snapshots__/index.spec.tsx.snap
+++ b/src/__tests__/__snapshots__/index.spec.tsx.snap
@@ -349,8 +349,8 @@ exports[`HTML Output given anyOf combiner placed next to allOf given allOf mergi
                   </div>
                   <div>
                     <span>Allowed values:</span>
-                    <span>\\"admin\\"</span>
-                    <span>\\"editor\\"</span>
+                    <span>admin</span>
+                    <span>editor</span>
                   </div>
                 </div>
                 <div></div>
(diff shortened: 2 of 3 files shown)


src/components/shared/Validations.tsx
```diff
@@ -63,11 +63,11 @@ const createValidationsFormatter = (name: string, options?: { exact?: boolean; n
 };
 
 const validationFormatters: Record<string, (value: unknown) => ValidationFormat | null> = {
-  enum: createValidationsFormatter('Allowed'),
-  examples: createValidationsFormatter('Example'),
+  enum: createValidationsFormatter('Allowed', { nowrap: true }),
```
P0lip, line 66: Hmm, and how will one recognized
```
enum: [1, "1"]
```
I don't think it's a good decision to remove quotes. These are strings, so let them be.

mallachari: We can't in such a case. That's why I added it in the first place.
@mnaumanali94 is it really needed to remove those quotes? We're losing a type then.

wmhilton: > Hmm, and how will one recognize such an enum?
> 
> ```
> enum: [1, "1"]
> ```

This is a terrible enum! If someone is mixing stringified numbers and numberified numbers in an enum, their API is beyond our help. They're a disgrace to the industry and we should encourage them to use SwaggerHub to hurt SwaggerHub's reputation. :trollface:

P0lip: I know it is, but it's just to show the example.
The below cases are valid tho:

```
enum: ["1", "2", "3"]
```

```
enum: [1, 2, 3]
```

If we render them the same way, how will one be able to tell the difference?

mpodlasin: @P0lip are enums the only potential problem? Or can you think about more examples where this change would cause confusion? 🤔 

What if we made the enums an exception here? @mnaumanali94 is a strong proponent of that change. 🤷

P0lip: Well, mostly enums since other keywords are not used for validation. 
I'd still keep it as is, though.
IMHO in this particular case we shouldn't prioritize a negligible benefit of improved readability over the potential of causing confusion and loss of valuable information.

I'll leave it to you all, tho.

P0lip: One idea I have in my mind is to make use of colors we use for types. Each example could have a different color based on its type.
That would satisfy all parties

mnaumanali94: The type of enums i.e. String vs Integer should show that information @P0lip or am i missing something?

P0lip: It depends on the scenario.
Consider the below:

```json
{ 
  "enum": ["0", 1, "2"]
}
```

or

```json
{
  "type": ["string", "number"],
  "enum": ["0", 1, "2"]
}
```

One won't be able to tell the difference once we apply the change.

mnaumanali94: Hmm. Have you ever seen an enum like this though? 🤦🏼

mnaumanali94: Should 
```
type: string,
enum: [asc, desc]
```

In our docs, these values should appear as "asc", "desc" is what you're suggesting?

P0lip: The first example is quite common actually.

mnaumanali94: I'm thinking this is not a common use case. The concern is primarily folks thinking `""` are part of the value which would be wrong. Lets get this in, and see if there's a reaction from our consumers. If so, we can use colours to denote string vs integer values.

mnaumanali94 (approved):

mmiask (approved):

PR #163, opened 2021-08-19
fix: check for array child enum

domagojk:
For stoplightio/elements#1665

I believe the issue occurs because type: array node is not recursively processed like object type so enum property was never found. This fixes the issue described in the ticket but it also changed the jest snapshot for a schema defined in `src/__fixtures__/tickets.schema.json` and indeed, found enums that should be included for a property like this:

```json
"availableDeliveryTypes": {
  "type": "array",
  "items": {
    "type": "string",
    "enum": [
      "POSTAL",
      "PICK_UP_STATION",
      "E-MAIL",
      "LOYALTY_CARD"
    ]
  }
}
```

### DIFF
--- a/src/components/shared/Validations.tsx
+++ b/src/components/shared/Validations.tsx
@@ -1,4 +1,4 @@
-import { RegularNode } from '@stoplight/json-schema-tree';
+import { isRegularNode, RegularNode } from '@stoplight/json-schema-tree';
 import { Flex, Text } from '@stoplight/mosaic';
 import { Dictionary } from '@stoplight/types';
 import capitalize from 'lodash/capitalize.js';
@@ -202,7 +202,15 @@ export function validationCount(schemaNode: RegularNode) {
 
 export function getValidationsFromSchema(schemaNode: RegularNode) {
   return {
-    ...(schemaNode.enum !== null ? { enum: schemaNode.enum } : null),
+    ...(schemaNode.enum !== null
+      ? { enum: schemaNode.enum }
+      : // in case schemaNode is type: "array", check if its child have defined enum
+      schemaNode.primaryType === 'array' &&
+        schemaNode.children?.length === 1 &&
+        isRegularNode(schemaNode.children[0]) &&
+        schemaNode.children[0].enum !== null
+      ? { enum: schemaNode.children[0].enum }
+      : null),
     ...('annotations' in schemaNode
       ? {
           ...(schemaNode.annotations.default ? { default: schemaNode.annotations.default } : null),
--- a/src/__tests__/__snapshots__/index.spec.tsx.snap
+++ b/src/__tests__/__snapshots__/index.spec.tsx.snap
@@ -3617,6 +3617,13 @@ exports[`HTML Output should match tickets.schema.json 1`] = `
                                     </div>
                                   </div>
                                 </div>
+                                <div>
+                                  <span>Allowed values:</span>
+                                  <span>\\"POSTAL\\"</span>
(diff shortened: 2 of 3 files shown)


src/components/shared/Validations.tsx
```diff
@@ -202,7 +202,15 @@ export function validationCount(schemaNode: RegularNode) {
 
 export function getValidationsFromSchema(schemaNode: RegularNode) {
   return {
-    ...(schemaNode.enum !== null ? { enum: schemaNode.enum } : null),
+    ...(schemaNode.enum !== null
+      ? { enum: schemaNode.enum }
+      : // in case schemaNode is type: "array", check if its child have defined enum
+      schemaNode.primaryType === 'array' &&
+        schemaNode.children?.length === 1 &&
```
P0lip, line 209: tgHmmm, but arrays can also have enums at the top, i.e.

```json
{
  "type": "array",
  "enum": [
    [0, 1],
    [2, 3]
  ],
  "items": {
    "type": "number"
  }
}
```

plus, there's also another "tuples" scenario possible:

```json
{
  "type": "array",
  "items": [
    {
      "type": "number",
      "enum": [0, 1, 2]
    },
    {
      "type": "string",
      "enum": ["a", "b", "c"]
    }
  ]
}
```

P0lip (commented):
Overall, arrays are troublesome to handle due to the way we represent them in the UI :/
I'll the final decision to you.

src/components/shared/Validations.tsx
domagojk: I knew you will find some cases I overlooked 😄 

If enum is defined for an array, validations will take take that one over child ones, so I think that solved the first thing, right?

I will check what do to with stuff like "touples", right now it wouldn't be supported since I only do this if there is a single item in array

P0lip: > If enum is defined for an array, validations will take take that one over child ones, so I think that solves the first thing, right?

The tricky part is that we should try to display both, because both enums impose certain restrictions.

Example:

```json
{
  "type": "array",
  "enum": [
    ["Maribor", "Zagreb", "Banja Luka", "Skopje"],
    ["Warsaw", "Prague", "Budapest", "Bratislava"]
  ],
  "items": {
    "type": "string",
    "enum": ["Warsaw", "Zagreb", "Prague", "Budapest", "Bratislava", "Maribor"]
  }
}
```

Now, this is valid `["Warsaw", "Prague", "Budapest", "Bratislava"]`, and this is not valid `["Maribor", "Zagreb", "Banja Luka", "Skopje"]`.
The former is valid because all constraints imposed by 2 enums are met, while the latter is invalid because the second enum restricts a certain set of cities.

Does that make sense?

In general, there's no good way to solve this issue. 
I'm just trying to point out the potential problem we might see.

I'm quite convinced that what you implemented here will be good enough for most cases, so we might just take it and merge it.

domagojk: Gotcha, well I would then go with current solution. It is not perfect but neither is what is in prod now. We can then revisit more complex enum scenarios later, if needed.

I'm adding @mnaumanali94 as a reviewer as well, so it is not only my decision 😄

mnaumanali94: Agreed. Lets get this in and as customers find/report complex cases we'll look into it.

mnaumanali94 (approved):

domagojk:
@P0lip having in mind that we agreed to release this non-perfect version, could you approve it? (I can't merge it now 😢 )

P0lip (approved):

PR #170, opened 2021-11-04
Add more UI feedback when JSON schema uses anyOf or oneOf

Nezteb:
Addresses https://github.com/stoplightio/platform-internal/issues/7717

Top level:
Screen Shot 2021-11-03 at 7 04 01 PM

Row level:
Screen Shot 2021-11-03 at 7 05 56 PM

### DIFF
--- a/src/components/SchemaRow/SchemaRow.tsx
+++ b/src/components/SchemaRow/SchemaRow.tsx
@@ -9,7 +9,6 @@ import {
 } from '@stoplight/json-schema-tree';
 import { Icon, Select } from '@stoplight/mosaic';
 import cn from 'classnames';
-import { first } from 'lodash';
 import last from 'lodash/last.js';
 import * as React from 'react';
 
--- a/src/components/SchemaRow/SchemaRow.tsx
+++ b/src/components/SchemaRow/SchemaRow.tsx
@@ -55,7 +55,7 @@ export const SchemaRow: React.FunctionComponent<SchemaRowProps> = ({ schemaNode,
   const isBrokenRef = typeof refNode?.error === 'string';
 
   const childNodes = React.useMemo(() => calculateChildrenToShow(typeToShow), [typeToShow]);
-  const combiner = anyOfOrOneOf(schemaNode);
+  const combiner = isRegularNode(schemaNode) && schemaNode.combiners?.length ? schemaNode.combiners[0] : null;
   return (
     <div className="sl-relative">
       <div className="sl-flex sl-max-w-full">
@@ -111,9 +111,7 @@ export const SchemaRow: React.FunctionComponent<SchemaRowProps> = ({ schemaNode,
                   />
                 )}
 
-                {schemaNode.subpath.length > 0 && combiner !== undefined && (
-                  <div className="sl-ml-1 sl-text-muted">{combiner}</div>
-                )}
+                {combiner !== null ? <div className="sl-ml-1 sl-text-muted">{combiner}</div> : null}
               </div>
               <Properties
                 required={isPropertyRequired(schemaNode)}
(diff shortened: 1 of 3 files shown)


src/components/SchemaRow/SchemaRow.tsx
```diff
@@ -9,6 +9,7 @@ import {
 } from '@stoplight/json-schema-tree';
 import { Icon, Select } from '@stoplight/mosaic';
 import cn from 'classnames';
+import { first } from 'lodash';
```
Nezteb, line 12: I need to get rid of this

P0lip:
what about `anyOf`? :D

Nezteb:
> what about `anyOf`? :D

Do you mean `allOf`? 😅

P0lip:
nah, there are 3 compound keywords - `oneOf`, `anyOf`, and `allOf` :D

Nezteb:
> nah, there are 3 compound keywords - `oneOf`, `anyOf`, and `allOf` :D

I took care of `anyOf` and `oneOf` (`allOf` wasn't listed in the ticket, I assume because it's... self-explanatory?), or did I miss something?

P0lip:
ah fuck, I'm blind, yes. You're right. I read `anyOf` as `allOf`.

src/components/SchemaRow/SchemaRow.tsx
```diff
@@ -55,6 +55,7 @@ export const SchemaRow: React.FunctionComponent<SchemaRowProps> = ({ schemaNode,
   const isBrokenRef = typeof refNode?.error === 'string';
 
   const childNodes = React.useMemo(() => calculateChildrenToShow(typeToShow), [typeToShow]);
+  const combiner = anyOfOrOneOf(schemaNode);
```
P0lip, line 58: A node should have a combiner property you could use
https://github.com/stoplightio/json-schema-tree/blob/master/src/nodes/RegularNode.ts#L40

```suggestion
  const combiners = isRegularNode(schemaNode) && schemaNode.combiners?.length ? schemaNode.combiners[0]: null;
```
I haven't tested it, but I believe it should work fine.

src/components/SchemaRow/TopLevelSchemaRow.tsx
```diff
@@ -23,6 +23,14 @@ export const TopLevelSchemaRow: React.FC<SchemaRowProps> = ({ schemaNode, nestin
   }
 
   if (isRegularNode(schemaNode) && choices.length > 1) {
+    let combinerType = '';
+    if (selectedChoice.type.subpath[0] === 'oneOf') {
```
P0lip, line 27: prob the same as above - could use `combiners` to determine it, but using `subpath` is good too.

P0lip (commented):
Looks good :+1:
You may want to use `combiners` tho if it happens to be yielding the results you expect.

src/components/SchemaRow/SchemaRow.tsx
Nezteb: Ah, my editor wasn't giving me that in auto-complete, I assume because I wasn't using `isRegularNode`. TIL!
Screen Shot 2021-11-04 at 11 41 21 AM

src/components/SchemaRow/SchemaRow.tsx
```diff
@@ -149,3 +152,15 @@ function shouldShowPropertyName(schemaNode: SchemaNode) {
...
+  if (schemaNode.fragment.anyOf !== undefined) {
+    return '(Any of)';
+  }
+
+  if (schemaNode.fragment.oneOf !== undefined) {
+    return '(One of)';
+  }
+
+  return undefined;
+}
```
P0lip, line 166: dead code now, nah? 
```suggestion
function anyOfOrOneOf(schemaNode: SchemaNode) {
  if (schemaNode.fragment.anyOf !== undefined) {
    return '(Any of)';
  }

  if (schemaNode.fragment.oneOf !== undefined) {
    return '(One of)';
  }

  return undefined;
}
```

or did we intend to use it in `src/components/SchemaRow/TopLevelSchemaRow.tsx`?

P0lip (approved):
:+1: just either remove `anyOfOrOneOf` fn or make use of it if needed

src/components/SchemaRow/SchemaRow.tsx
Nezteb: :ope: my bad

jgreen44 (approved):

P0lip:
@Nezteb you got to update snapshots :D

PR #175, opened 2021-11-23
chore: use mosaic consistently

lottamus:
Element's HttpOperation Request parameters component is using Mosaic's spacing but JSV is not. This results in a 1px difference between the parameters and the body section. This hurts my eyeballs. 🙈 

https://github.com/stoplightio/elements/blob/350e46ae4facb93c42997dd97cd57be98b947b88/packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx#L38

Screen Shot 2021-11-22 at 9 39 02 PM

Screen Shot 2021-11-22 at 9 38 51 PM

Thus I took it upon myself to use Mosaic consistently within this lib.


P0lip:
We didn't use mosaic for a reason.
https://github.com/stoplightio/json-schema-viewer/pull/113 
Have you confirmed the perf issues are no longer apparent?

marbemac:
FWIW, on my macbook rendering the stress test storybook:

**master**

`defaultExpandedDepth = 2`, around ~95ms:

Screen Shot 2021-11-24 at 12 10 56 AM

`defaultExpandedDepth = 5`, around ~460ms:

Screen Shot 2021-11-24 at 12 11 33 AM

**this PR**

`defaultExpandedDepth = 2`, around ~145ms (1.5x):

Screen Shot 2021-11-24 at 12 12 26 AM

`defaultExpandedDepth = 5`, around ~870ms (1.9x):

Screen Shot 2021-11-24 at 12 12 55 AM

---

IMHO the stress test example with expanded depth of 5 is not a use case we need to serve 🤷‍♂️. We don't ever (or should not) set default expand depth to larger than 2, which means realistically we'll never be rendering anything close to the second example above. 

95ms versus 145ms for a stress test style use case is not a huge deal IMHO.

P0lip:
> 145ms versus 200ms for a stress test style use case is not a huge deal IMHO.

I will admit that to me (1.37x) slowdown is a notable one, particularly if we speak of tens or hundreds of ms and take into account the fact we usually don't render JSV alone, but a whole docs page that has its own set of accompanying components.
The UI library shouldn't introduce any meaningful overhead and this is clearly not the case here. 
I'll leave the call to you all, but I'd say we shouldn't keep the status quo and not change anything.

60ms equals almost 4 frames.

marbemac:
What's in place now just isn't really a good option - I guess we could look for another way that doesn't involve mosaic, buttt.. 🤷‍♂️. JSV IS using mosaic right now, just a "private" API from mosaic (hardcoded classnames that come from mosaic css).

Check this out though:

- Some adjustments to the JSV to render a few less wrapper elements that were not needed
- https://github.com/stoplightio/mosaic/pull/313 (as I said a million times, I am/was almost certain that "Box" / "forwardRef" were not the culprits)

With the two changes above (I yalc'd that mosaic PR into JSV), the perf is basically the same as on main branch:

Screen Shot 2021-11-25 at 1 48 09 PM

P0lip (approved):

marbemac (approved):

Nezteb (approved):
With the new mosaic memoization this is 👌.

PR #178, opened 2021-11-29
fix: removed NameValidations to properly show default values

domagojk:
For stoplightio/platform-internal#9043

I couldn't find the reason why `NameValidations` were used for boolean types so I removed it 🤷‍♂️ 

Now, in case of

```json
"defaultExample1": {
  "type": "boolean",
  "default": true
},
"defaultExample2": {
  "type": "boolean",
  "default": false
}
```

It shows:

Screenshot 2021-11-29 at 15 02 44

Rather than displaying `Default` in case `default: true` and nothing in case of `default: false`.

Besides that, I noticed it also fixed the issue with some other boolean types like:

```json
"something": {
  "type": "boolean",
  "example": true
}
```

Screenshot 2021-11-29 at 15 04 25

Which is was previously displayed as:

Screenshot 2021-11-29 at 15 06 02

@mallachari Do you remember why was `NameValidations` needed for booleans? Is there some cases that i'm not seeing which this PR could broke?

### DIFF
--- a/src/components/shared/Validations.tsx
+++ b/src/components/shared/Validations.tsx
@@ -5,7 +5,6 @@ import capitalize from 'lodash/capitalize.js';
 import keys from 'lodash/keys.js';
 import omit from 'lodash/omit.js';
 import pick from 'lodash/pick.js';
-import pickBy from 'lodash/pickBy.js';
 import uniq from 'lodash/uniq.js';
 import * as React from 'react';
 
@@ -108,13 +107,8 @@ function filterOutOasFormatValidations(format: string, values: Dictionary<unknow
 
 export const Validations: React.FunctionComponent<IValidations> = ({ validations, hideExamples }) => {
   const numberValidations = pick(validations, numberValidationNames);
-  const booleanValidations = omit(
-    pickBy(validations, v => ['true', 'false'].includes(String(v))),
-    excludedValidations,
-  );
   const keyValueValidations = omit(validations, [
     ...keys(numberValidations),
-    ...keys(booleanValidations),
     ...excludedValidations,
     ...(hideExamples ? exampleValidationNames : []),
   ]);
@@ -123,7 +117,6 @@ export const Validations: React.FunctionComponent<IValidations> = ({ validations
     <>
       <NumberValidations validations={numberValidations} />
       <KeyValueValidations validations={keyValueValidations} />
-      <NameValidations validations={booleanValidations} />
     </>
   );
 };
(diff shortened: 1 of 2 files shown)


P0lip:
>  boolean types 

it's not boolean types, but boolean-ish validations, like, say

```yaml
exclusiveMinimum: true
```

I know it's not the best example since it's draft 4, but hopefully you get the idea.

src/components/shared/Validations.tsx
```diff
@@ -108,13 +107,8 @@ function filterOutOasFormatValidations(format: string, values: Dictionary<unknow
 
 export const Validations: React.FunctionComponent<IValidations> = ({ validations, hideExamples }) => {
   const numberValidations = pick(validations, numberValidationNames);
-  const booleanValidations = omit(
```
P0lip, line 111: you prob want to revert it

domagojk: that excludes it both from `booleanValidations` and `keyValueValidations` but I want them to still show in keyVal ones. But I can fix that. I'm just wondering, Is there any usecase where it makes more sense to use `NameValidations`  rather than `KeyValueValidations`?

Because, if the value is `true` or `false` I think it makes more sense to display it using `KeyValueValidations`, so it says `default: true` rather than `Default`. Same for `example` or even for `exclusiveMinimum` (which wont be shown here because it falls in `numberValidations`)

P0lip: > Is there any usecase where it makes more sense to use NameValidations rather than KeyValueValidations?

That's a good question... I'm not really sure whether we actually have any booleanish validations left now that we filter out default, readOnly, writeOnly, and everything related to numbers is handled separately. 

>  I think it makes more sense to display it using KeyValueValidations, so it says default: true rather than Default

I agree

mallachari: We used NameValidations before because some of they sounded better this way (the property name was enough) it ie. deprecated, readOnly, writeOnly etc. Default can be any type, but agree that as boolean it might look weird. If it makes more sense to change it given those validations are represented differently now then go ahead, I think it's fine.

domagojk: Gotcha. Yea I think most of them are handled differently now (all those mentioned included)

domagojk:
Ok @P0lip @mallachari then my suggestion is to leave the PR as is, which means, in case we missed any "booleanish validation" it will just be represented with `KeyValueValidations` rather than `NameValidations`. 

And in case there is some we are missing, we can bring it back, but I would do it explicitly this time (not by default like now)

Approve if you agree 😄

mnaumanali94 (approved):

P0lip (approved):
ajmo

mallachari (approved):

PR #179, opened 2021-11-30
feat: tree-ui and path crumbs

marbemac:
closes https://github.com/stoplightio/platform-internal/issues/9189

Apologies in advance for the large PR - if it's too much I can take some time to try and break it out into 2 or 3 PRs just LMK.

- remove some dom complexity that seemed unnecessary
- - a couple of new "real world" examples in the storybook (box file and github issue)... please note that ~2200 of the added lines come from just this
- a lot of the diff also comes from snapshot updates
- new tree UI to better visualize nesting, and some polish on indentation / vertical alignment of a variety of cases
- new path crumbs to help the consumer understand where a given property lives in the schema when the schema is large or deeply nested
- some small optimizations to cleanup spacing around validations
- "show more" button if a description has more than one paragraph
- new option: `renderRootTreeLines`
- new option: `disableCrumbs`
- new option: `maxHeight`

---

**Tree "guidelines" + crumbs:**

Screen Shot 2021-11-29 at 7 05 34 PM

---

Small adjustments to the validations layout, including a fix such that long validations (like the long example in the first row) do not push the "custom row addon" off the page. Note how in the first image, the checkbox on the right side is pushed outside of the JSV container.

**Before**

Screen Shot 2021-11-30 at 3 57 43 PM

**After**

Screen Shot 2021-11-30 at 3 56 47 PM

---

Dark mode still works nicely:

Screen Shot 2021-11-30 at 4 58 26 PM


src/components/SchemaRow/TopLevelSchemaRow.tsx
```diff
@@ -1,85 +1,106 @@
...
+
+        {childNodes.length > 0 ? <ChildStack childNodes={childNodes} currentNestingLevel={nestingLevel} /> : null}
+      </>
     );
   }
 
-  return <SchemaRow schemaNode={schemaNode} nestingLevel={nestingLevel} />;
+  return (
+    <>
+      <ScrollCheck />
```
P0lip, line 86: Unless something has changed, I recall `IntersectionObserver` a pretty bad impact on perf when you had thousands of instances.
This may be not the case here due to a smaller quantity of nodes, but it's certainly something to watch out for.
I'll take a closer look at it later today

package.json
```diff
@@ -38,31 +38,32 @@
...
   },
   "dependencies": {
     "@fortawesome/free-solid-svg-icons": "^5.15.2",
-    "@stoplight/json": "^3.10.0",
+    "@stoplight/json": "^3.17.1",
     "@stoplight/json-schema-tree": "^2.1.2",
     "@stoplight/react-error-boundary": "^2.0.0",
     "@types/json-schema": "^7.0.7",
     "classnames": "^2.2.6",
+    "jotai": "^1.4.5",
```
P0lip, line 54: meh, do we really need yet another library?
I know we probably use it somewhere in our ecosystem, but given it's a library meant to be consumed by others that may not have it installed it'd be better to rely on contexts.
Is there any reason we cannot use contexts here?

src/components/PathCrumbs/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+      color="light"
+      alignItems="center"
+    >
+      {parentCrumbElems}
+      {pathCrumbElems.length && <HStack divider={<Box fontWeight="bold">.</Box>}>{pathCrumbElems}</HStack>}
+    </HStack>
+  );
+};
+
+function propertyPathToObjectPath(node: SchemaNode) {
```
P0lip, line 71: @marbemac I've changed it a bit.
We could actually use titles if we wanted to.

P0lip (commented):
Some early feedback

src/components/SchemaRow/TopLevelSchemaRow.tsx
marbemac: There's only one instance though?

package.json
marbemac: Yeah we have jotai in the rest of our ecosystem, and it's 2.6kb gzipped 🤷‍♂️. I had thought about context, but it adds then need to start bringing in React.useReducer and it becomes a larger change to the options context object we already have (at least how I think about mutable react context, there is prob a way to keep things almost the same). I was also hoping to improve the expanded/collapsed state in a future PR so that it remembers which children are expanded/collapsed when a parent is toggled, and figured could use something like jotai to help with that as well. What do you think?

src/components/PathCrumbs/index.tsx
marbemac: Cool I'll check it out! Re titles that would change the intent of these crumbs. They are meant to allow you to easily understand where a given property sits within the larger object structure - using titles wouldn't achieve that, or maybe I misunderstand?

src/components/SchemaRow/TopLevelSchemaRow.tsx
P0lip: Per each row no?

src/components/PathCrumbs/index.tsx
marbemac: Couple of missed bits - root array should just be `[]` (rather than `undefined[]`):

Screen Shot 2021-11-30 at 10 28 56 AM

Do we want the leading `.` (my original version did not have this). it might confuse people?:

Screen Shot 2021-11-30 at 10 30 05 AM

src/components/SchemaRow/TopLevelSchemaRow.tsx
marbemac: Just for the `TopLevelSchemaRow`, which is only rendered once at the top

src/components/PathCrumbs/index.tsx
P0lip: I meant something like `title / title / title`, assuming they are available

src/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -4,39 +4,27 @@ exports[`Expanded depth nested object static given initial level set to 0, shoul
 "<div class=\\"\\" id=\\"mosaic-provider-react-aria-0-1\\">
```
P0lip, line 4: at this point it feels appropriate to remove the snapshots.
Let me explain why I added them in the first place.
About a year ago I was moving the whole logic off to JST. This is the PR that introduced the following snapshots https://github.com/stoplightio/json-schema-viewer/pull/92.
Once JST was founded, I incorporated it over here https://github.com/stoplightio/json-schema-viewer/pull/91/files and these snapshots were very helpful and ensured me that JST didn't have any meaningful impact on the UI.
Besides a few new tests and actual bug fixes (JST did help to fix a number of bugs as well as improved the overall processing of schemas), the snapshots were left untouched.


Now, when we actively work on UI, they don't have any value.

src/components/SchemaRow/TopLevelSchemaRow.tsx
P0lip: ah yeah, true, me blind again.

src/__tests__/__snapshots__/index.spec.tsx.snap
P0lip: done here https://github.com/stoplightio/json-schema-viewer/pull/180/files

src/consts.ts
```diff
@@ -8,10 +8,12 @@ export const COMBINER_PRETTY_NAMES: Readonly<Dictionary<string, SchemaCombinerNa
...
-export const CARET_OFFSET = NEGATIVE_NESTING_OFFSET;
-export const CARET_ICON_SIZE = 'xs';
-export const CARET_ICON_BOX_DIMENSION = 3;
+export const CARET_ICON_SIZE = 'sm';
+
+export const COMBINER_NAME_MAP: Dictionary<string, SchemaCombinerName> = {
+  allOf: 'all of',
+  anyOf: 'any of',
+  oneOf: 'one of',
+};
```
Nezteb, line 19: 👌

src/consts.ts
```diff
@@ -8,10 +8,12 @@ export const COMBINER_PRETTY_NAMES: Readonly<Dictionary<string, SchemaCombinerNa
...
-// @ts-expect-error: negative offset
-export const NEGATIVE_NESTING_OFFSET: NegativeSpaceVals = -NESTING_OFFSET;
+export const NESTING_OFFSET: SpaceVals = 3;
 
-export const CARET_OFFSET = NEGATIVE_NESTING_OFFSET;
-export const CARET_ICON_SIZE = 'xs';
-export const CARET_ICON_BOX_DIMENSION = 3;
+export const CARET_ICON_SIZE = 'sm';
+
+export const COMBINER_NAME_MAP: Dictionary<string, SchemaCombinerName> = {
```
Nezteb, line 15: At first I was confused because I'd assume a `Dictionary` type's generics would be the key first and then the value, but apparently our `Dictionary` is backwards:
```
type Dictionary<T, K extends string | number = string> = { [key in K]: T };
```

So that makes sense. 😂

src/components/SchemaRow/TopLevelSchemaRow.tsx
```diff
@@ -1,85 +1,107 @@
...
               </Pressable>
             )}
           />
 
-          {childNodes.length > 0 ? <ChildStack childNodes={childNodes} currentNestingLevel={nestingLevel} /> : null}
-        </Box>
-      </DecreaseIndentation>
+          {combiner !== null ? (
+            <Flex alignItems="center" color="muted" fontSize="base">
+              {`(${COMBINER_NAME_MAP[combiner]})`}
```
Nezteb, line 61: At first I thought this wasn't working for some schemas:

Working:
Screen Shot 2021-11-30 at 4 50 20 PM

Not working:
Screen Shot 2021-11-30 at 4 50 30 PM

But then I realized that although the schema has a `oneOf` key, it's nested in the `items` object. 🤦‍♂️ https://github.com/stoplightio/json-schema-viewer/blob/3ceb51e692f709784774964cab6d119ebd38a235/src/__fixtures__/combiners/oneof-within-array-item.json#L3-L5

I'm not sure what should happen in this case. Is the schema wrong/invalid? Or do we just not show the combiner name for this schema and call it good?

marbemac: Hmm I think this has been an issue on main branch (not introduced by this PR) - can fix it in this PR or we can follow-up in a separate PR. What do you think?

Nezteb: I'm fine with a followup PR!

src/components/SchemaRow/SchemaRow.tsx
```diff
@@ -52,89 +47,94 @@ export const SchemaRow: React.FunctionComponent<SchemaRowProps> = ({ schemaNode,
 
   const isBrokenRef = typeof refNode?.error === 'string';
 
+  const rootLevel = renderRootTreeLines ? 1 : 2;
   const childNodes = React.useMemo(() => calculateChildrenToShow(typeToShow), [typeToShow]);
   const combiner = isRegularNode(schemaNode) && schemaNode.combiners?.length ? schemaNode.combiners[0] : null;
+  const isCollapsible = childNodes.length > 0;
+  const isRootLevel = nestingLevel < rootLevel;
```
Nezteb, line 54: I don't know if it matters, but should this be `<=`?

src/hooks/useOnScreen.ts
```diff
@@ -0,0 +1,41 @@
+import { RefObject, useEffect, useRef, useState } from 'react';
+
+export const useOnScreen = (ref: RefObject<HTMLElement>) => {
```
Nezteb, line 3: Nitpick: Should this be `useIsOnScreen`?

Nezteb (approved):
I like this a lot. It's incredibly clean. My comments are mostly nitpicks that probably don't affect anything, so I'm fine approving without any changes. 🚀

src/hooks/useOnScreen.ts
marbemac: yes nice catch - done!

src/components/SchemaRow/SchemaRow.tsx
marbemac: Not here actually - yeah it's kind of confusing 🤔

Nezteb: That's fine I just wasn't sure if `nestingLevel === rootLevel` counted as `isRootLevel` or not. 😄

src/hooks/useIsOnScreen.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  useEffect(() => {
+    if (ref.current) {
+      const scrollParent = getScrollParent(ref.current);
+      observerRef.current = new IntersectionObserver(
+        ([entry]) => {
+          setIsOnScreen(entry.isIntersecting);
+        },
+        {
+          root: scrollParent === window ? undefined : (scrollParent as HTMLElement),
```
P0lip, line 15: won't this throw an exception in the SSR world?
`window` is oftentimes not defined there.

src/__stories__/utils/Wrapper.tsx
```diff
@@ -2,19 +2,17 @@ import * as React from 'react';
 
 export const Wrapper: React.FunctionComponent<React.HTMLAttributes<HTMLElement>> = ({ children, ...props }) => {
```
P0lip, line 3: We should be able to remove it now. I forgot to do it in https://github.com/stoplightio/json-schema-viewer/pull/179/commits/95cac3c2332cf8cb839b5418548a4a9de84d9d56

src/hooks/useIsOnScreen.ts
P0lip:
```suggestion
          root: typeof window === 'undefined' || scrollParent === window ? undefined : (scrollParent as HTMLElement),
```
or
```suggestion
          root: scrollParent === globalThis.window ? undefined : (scrollParent as HTMLElement),
```
albeit I'm not sure whether the latter is supported well in older environments. Presumably not, so the first one is the safe bet

src/hooks/useIsOnScreen.ts
```diff
@@ -0,0 +1,41 @@
...
+          rootMargin: '200px 0px 0px 0px',
```
P0lip, line 16: shouldn't we use a viewport unit? Example:
```suggestion
          rootMargin: '20vh 0px 0px 0px',
```

src/components/shared/__tests__/Property.spec.tsx
```diff
@@ -141,7 +141,7 @@ describe('Property component', () => {
...
...ed=\\"false\\" id=\\"react-aria-0-4\\" aria-labelledby=\\"react-aria-0-4 react-aria-0-6\\" type=\\"button\\" class=\\"sl-button sl-w-full sl-h-sm sl-text-base sl-font-normal sl-px-1.5 sl-bg-transparent sl-rounded sl-border-transparent hover:sl-border-input focus:sl-border-primary active:sl-border-primary sl-border disabled:sl-opacity-60\\"><div class=\\"sl-flex sl-flex-1 sl-justify-items-start sl-items-center\\"><div class=\\"sl-pr-1\\">array[strings]</div></div><div class=\\"sl-text-xs sl--mr-0.5 sl-ml-1\\"><div class=\\"sl-pt-0.5 sl-pr-0.5\\"><svg aria-hidden=\\"true\\" focusable=\\"false\\" data-prefix=\\"fas\\" data-icon=\\"chevron-down\\" class=\\"svg-inline--fa fa-chevron-down fa-w-14 fa-xs sl-icon\\" role=\\"img\\" xmlns=\\"http://www.w3.org/2000/svg\\" viewBox=\\"0 0 448 512\\"><path fill=\\"currentColor\\" d=\\"M207.029 381.476L12.686 187.132c-9.373-9.373-9.373-24.569 0-33.941l22.667-22.667c9.357-9.357 24.522-9.375 33.901-.04L224 284.505l154.745-154.021c9.379-9.335 24.544-9.317 33.901.04l22.667 22.667c9.373 9.373 9.373 24.569 0 33.941L240.971 381.476c-9.373 9.372-24.569 9.372-33.942 0z\\"></path></svg></div></div></button></div></div></div></div></div></div></div></div>"`,
```
P0lip, line 144: ah yeah, that's a good change, I like it. :trollface:

src/components/shared/Validations.tsx
```diff
@@ -131,13 +131,13 @@ const NumberValidations = ({
     return null;
   }
   return (
-    <Flex color="muted" maxW="full">
+    <HStack color="muted" maxW="full" spacing={1}>
       {entries
         .map(([key, value]) => numberValidationFormatters[key](value))
         .map((value, i) => (
```
P0lip, line 137: feels redundant, but since it's an existing code and this PR is already fairly large, let's keep it unchanged

src/components/shared/Validations.tsx
```diff
@@ -54,19 +54,19 @@ const createValidationsFormatter = (name: string, options?: { exact?: boolean; n
...
+      name: options?.exact ? name : values.length > 1 ? `${name}s` : `${name}`,
       values: values.map(createStringFormatter(options?.nowrap)),
     };
   }
   return null;
 };
 
 const validationFormatters: Record<string, (value: unknown) => ValidationFormat | null> = {
-  enum: createValidationsFormatter('Allowed', { nowrap: true }),
+  enum: createValidationsFormatter('Allowed value', { nowrap: true }),
```
P0lip, line 65:
```suggestion
  enum: createValidationsFormatter('Allowed values', { nowrap: true }),
```
shouldn't this say "values"?

src/components/shared/Description.tsx
```diff
@@ -1,13 +1,47 @@
...
+    <Box
+      as={MarkdownViewer}
+      markdown={firstParagraph}
+      parseOptions={{
+        components: {
+          p: (props: any) => {
+            return (
+              <Box as="p">
+                <Text mr={1}>{props.children}</Text>
+                <Link cursor="pointer" onClick={() => setShowAll(true)}>
```
P0lip, line 34: isn't this a button?

src/components/SchemaRow/TopLevelSchemaRow.tsx
```diff
@@ -1,85 +1,107 @@
...
+import { COMBINER_NAME_MAP } from '../../consts';
+import { useIsOnScreen } from '../../hooks/useIsOnScreen';
 import { calculateChildrenToShow, isComplexArray } from '../../tree';
+import { showPathCrumbsAtom } from '../PathCrumbs';
 import { ChildStack } from '../shared/ChildStack';
 import { SchemaRow, SchemaRowProps } from './SchemaRow';
 import { useChoices } from './useChoices';
 
-export const TopLevelSchemaRow: React.FC<SchemaRowProps> = ({ schemaNode, nestingLevel }) => {
+export const TopLevelSchemaRow = ({ schemaNode }: Pick<SchemaRowProps, 'schemaNode'>) => {
```
P0lip, line 15: is there any reason we dropped `React.FC`? It also ensures the return type is correct, while this one doesn't do it anymore which can potentially lead to invalid renders if undefined is returned.

src/hooks/useIsOnScreen.ts
marbemac: Nope because it's in a `useEffect` - `useEffect` only runs in browser

src/components/SchemaRow/SchemaRow.tsx
```diff
@@ -24,11 +20,10 @@ export interface SchemaRowProps {
...
 
 export const SchemaRow: React.FunctionComponent<SchemaRowProps> = ({ schemaNode, nestingLevel }) => {
-  const { defaultExpandedDepth, renderRowAddon, onGoToRef, hideExamples } = useJSVOptionsContext();
+  const { defaultExpandedDepth, renderRowAddon, onGoToRef, hideExamples, renderRootTreeLines } = useJSVOptionsContext();
 
-  const [isExpanded, setExpanded] = React.useState<boolean>(
-    !isMirroredNode(schemaNode) && nestingLevel <= defaultExpandedDepth,
-  );
+  const setPathCrumbs = useUpdateAtom(pathCrumbsAtom);
+  const [isExpanded, setExpanded] = React.useState<boolean>(nestingLevel <= defaultExpandedDepth);
```
P0lip, line 26: `isMirroredNode(schemaNode)` bit is super important, we need to add it back
```suggestion
  const [isExpanded, setExpanded] = React.useState<boolean>(!isMirroredNode(schemaNode) && nestingLevel <= defaultExpandedDepth);
```

We'll enter an infinite loop if `defaultExpandedDepth` is set to `Infinity` which actually might be the case somewhere in platform 😬.

If you do want them to be expanded, then perhaps let's add such a check

```suggestion
  const [isExpanded, setExpanded] = React.useState<boolean>((Number.isFinite(defaultExpandedDepth) || !isMirroredNode(schemaNode)) && nestingLevel <= defaultExpandedDepth);
```

P0lip (commented):
I'll probably need a second pass on it, but left a few comments

src/components/shared/Validations.tsx
marbemac: The `createValidationsFormatter` pluralizes it if needed

src/components/SchemaRow/TopLevelSchemaRow.tsx
marbemac: Yeah `React.FC` is generally over-used in a few places. It adds props to the typings such as `children` that are not actually supported or used by this component.

Re the return undefined thing - if you try adding a `return undefined` anywhere in here you'll see that the typings call it out wherever this component is used (such as the `JsonSchemaViewer` component).

src/components/SchemaRow/SchemaRow.tsx
marbemac: Let me remind myself why I did this change, one sec :).

src/components/shared/Description.tsx
marbemac: I could add `role="button"`, but link works just fine as a clickable and tabbable thing 🤷‍♂️. I don't feel strongly about it, but might be annoying to render a button here that doesn't have button styling.

src/components/SchemaRow/SchemaRow.tsx
marbemac: You're right, nice catch - fixed.

src/hooks/useIsOnScreen.ts
marbemac: Screen Shot 2021-12-01 at 4 33 45 PM

marbemac: Meant to lower to 100px though, doing that.

src/hooks/useIsOnScreen.ts
P0lip: Yeah, I wasn't sure whether `useEffect` would be run or not, but I suppose it's just the sync one so the `useLayoutEffect`

src/hooks/useIsOnScreen.ts
P0lip: yeah, percentages feel a bit weird. 100px works.

src/components/SchemaRow/SchemaRow.tsx
```diff
@@ -52,89 +56,94 @@ export const SchemaRow: React.FunctionComponent<SchemaRowProps> = ({ schemaNode,
 
   const isBrokenRef = typeof refNode?.error === 'string';
 
+  const rootLevel = renderRootTreeLines ? 1 : 2;
   const childNodes = React.useMemo(() => calculateChildrenToShow(typeToShow), [typeToShow]);
   const combiner = isRegularNode(schemaNode) && schemaNode.combiners?.length ? schemaNode.combiners[0] : null;
```
P0lip, line 61: this is wrong, but we can fix it some other day.
There might be multiple combiners :grimacing:, but I'm not sure how we'd like to render that.

src/components/SchemaRow/TopLevelSchemaRow.tsx
P0lip: > Yeah React.FC is generally over-used in a few places. It adds props to the typings such as children that are not actually supported or used by this component.

because `children` is something that's always supported? even if not consumed, you can still pass it around, it's just some sort of a native property supported by React :man_shrugging: 
I dunno, this feels just less correct than having `React.FC`, but it's a topic for another convo.

src/components/shared/Description.tsx
P0lip: If we want to be more correct semantically this would be a button, but I get a different styling would be applied, so meh.

src/hooks/useIsOnScreen.ts
```diff
@@ -0,0 +1,41 @@
+import { RefObject, useEffect, useRef, useState } from 'react';
+
+export const useIsOnScreen = (ref: RefObject<HTMLElement>) => {
+  const observerRef = useRef<IntersectionObserver | null>(null);
+  const [isOnScreen, setIsOnScreen] = useState(true);
+
+  useEffect(() => {
+    if (ref.current) {
+      const scrollParent = getScrollParent(ref.current);
+      observerRef.current = new IntersectionObserver(
```
P0lip, line 10:
```suggestion
      const observer = new IntersectionObserver(
```
we don't really need that `observerRef` do we? The intersection observer doesn't leak outside of the effect hook.

P0lip (approved):
:+1:

src/components/SchemaRow/TopLevelSchemaRow.tsx
marbemac: This component does not do anything with children - the consumer should not pass in children or think that they can and it would do something. It's like adding a random typing to your component that is not used at all

src/hooks/useIsOnScreen.ts
marbemac: Yup!

src/components/SchemaRow/SchemaRow.tsx
marbemac: Yesss I did notice a number of other weird things as I was going through this - but I already got pulled wayyy too far into the rabbit hole with the changes here already so 🤷‍♂️.

src/hooks/useIsOnScreen.ts
marbemac: good call, done

PR #180, opened 2021-11-30
chore: remove html output tests

P0lip:
https://github.com/stoplightio/json-schema-viewer/pull/179#discussion_r759670237

> at this point it feels appropriate to remove the snapshots.
Let me explain why I added them in the first place.
About a year ago I was moving the whole logic off to JST. This is the PR that introduced the following snapshots #92.
Once JST was founded, I incorporated it over here #91 and these snapshots were very helpful and ensured me that JST didn't have any meaningful impact on the UI.
Besides a few new tests and actual bug fixes (JST did help to fix a number of bugs as well as improved the overall processing of schemas), the snapshots were left untouched.

> Now, when we actively work on UI, they don't have any value.


marbemac (approved):

marbemac:
👍 can we remove code climate while we're at it? we don't use it anymore in our ecosystem

P0lip:
@marbemac we don't have any code related to CC in our repo.
I think it just works because it's an open-source project and we have some webhooks attached.
can prob remove it here https://codeclimate.com/repos/5c407ac419d5de0257011e41/edit I guess?

domagojk (approved):

PR #181, opened 2021-12-02
fix: path crumbs compat with virtual scroll containers

marbemac:
E.g. the one in legacy ui-kit that is used in Studio.

### DIFF
--- a/src/hooks/useIsOnScreen.ts
+++ b/src/hooks/useIsOnScreen.ts
@@ -34,7 +34,7 @@ function getScrollParent(node: HTMLElement | null): HTMLElement | typeof window 
     return null;
   }
 
-  if (node.scrollHeight > node.clientHeight) {
+  if (node.scrollHeight > node.clientHeight && node.clientHeight > 0) {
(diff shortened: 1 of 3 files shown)


src/hooks/useIsOnScreen.ts
```diff
@@ -34,7 +34,7 @@ function getScrollParent(node: HTMLElement | null): HTMLElement | typeof window
     return null;
   }
 
-  if (node.scrollHeight > node.clientHeight) {
+  if (node.scrollHeight > node.clientHeight && node.clientHeight > 0) {
```
marbemac, line 37: The ui-kit scroll container wraps things with a `0` height `0` width "scrollable" container (freaking weird lol), which breaks this behavior. Added this extra check to ignore 0 height parent containers, which fixes the issue (I yalc'd and confirmed this in platform).

Screen Shot 2021-12-02 at 4 42 23 PM

P0lip (approved):

wmhilton (approved):
I believe you.

PR #182, opened 2021-12-03
fix: show guideline to right props on hover if present

marbemac:
To help visually connect prop to extra values on right if there are any.


wmhilton:
screenshot looks good! gonna try running storybook locally. I'm curious how it handles on touch screens or with keyboard focus.

marbemac:
Good call! I'll check too

marbemac:
So obv no hover on touch screen, but will show on tap in this case 🤷‍♂️

marbemac:
On the other side of things, I have a whole lot of ideas re keyboard accessibility / navigation that I'd like to visit at some point.

wmhilton:
(Keyboard accessibility is a big ask obviously, but touch support is simple enough.)

Edit: Looks like `setHoveredNode` is called when nodes are touched, so looks nice on mobile 👍🏻

wmhilton (approved):
Looks beautiful!

PR #183, opened 2021-12-30
feat: show combiner for properties of an object

paulatulis:
Attempt to address #[9513](https://app.zenhub.com/workspaces/-team-pierogi-platoon-610871b8d6e5310013071b72/issues/stoplightio/platform-internal/9513)

Screen Shot 2021-12-30 at 9 22 35 AM

Because the combiner is a level deep, it shows up in each choice. Is that too busy for someone's eyes?


src/__stories__/JsonSchemaViewer.tsx
```diff
@@ -159,19 +160,26 @@ storiesOf('JsonSchemaViewer/combiners', module)
       onGoToRef={action('onGoToRef')}
     />
   ))
-  .add('anyOf-array-schema', () => (
+  .add('oneOf-array-schema', () => (
```
paulatulis, line 163: this change is unrelated to this issue, but I also updated the story names to match the files/schema

src/__stories__/JsonSchemaViewer.tsx
```diff
@@ -159,19 +160,26 @@ storiesOf('JsonSchemaViewer/combiners', module)
...
     <JsonSchemaViewer
       schema={oneOfWithArraySchema as JSONSchema4}
       defaultExpandedDepth={number('defaultExpandedDepth', 1)}
       onGoToRef={action('onGoToRef')}
     />
   ))
-  .add('anyOf-array-schema2', () => (
+  .add('oneOf-array-schema2', () => (
```
paulatulis, line 170: this change is unrelated to this issue, but I also updated the story names to match the files/schema

Nezteb (approved):
I tried this out locally and it looks good. 👌

mpodlasin:
Yeah, I would be curious to hear from Nauman what would he expect in the select options 😅 

If there is no array, you see only options in select, without anyOf/oneOf:

Screenshot 2022-01-04 at 15 12 18
Screenshot 2022-01-04 at 15 12 23

mnaumanali94:
> If there is no array, you see only options in select, without anyOf/oneOf:

@mpodlasin I thought we already covered the case for anyOf/OneOf showing for non array cases.

PR #205, opened 2022-08-23
fix: render top-level description

P0lip:
Fixes https://github.com/stoplightio/platform-internal/issues/12777


mmiask (approved):
Everything looks good

PR #206, opened 2022-09-14
feat: add skipTopLevelDescription option

marbemac:
Discussed with @P0lip on slack.

Follow-up to https://github.com/stoplightio/json-schema-viewer/pull/205 that adds a `skipTopLevelDescription` prop so that users can opt into old behavior (not rendering top level description).

We need this for standalone schema docs pages in elements.


P0lip (commented):
:+1:

PR #243, opened 2023-11-15
feat: pass maxRefDepth to tree to limit recursion

daniel-white:
upgrade json-schema-tree and pass the new option around

### DIFF
--- a/src/components/JsonSchemaViewer.tsx
+++ b/src/components/JsonSchemaViewer.tsx
@@ -23,6 +23,8 @@ export type JsonSchemaProps = Partial<JSVOptions> & {
   emptyText?: string;
   className?: string;
   resolveRef?: SchemaTreeRefDereferenceFn;
+  /** Controls the level of recursion of refs. Prevents overly complex trees and running out of stack depth. */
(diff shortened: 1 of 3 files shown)


src/components/JsonSchemaViewer.tsx
```diff
@@ -106,6 +110,7 @@ const JsonSchemaViewerInner = ({
     const jsonSchemaTree = new JsonSchemaTree(schema, {
       mergeAllOf: true,
       refResolver: resolveRef,
+      maxRefDepth,
```
EdVinyard, line 113: Is it correct that `JsonSchemaTree` already had an optional `maxRefDepth` parameter, and this PR just allows the caller to specify that when instantiating a `JsonSchemaViewer` component?  (And if it's not specified, there's no limit?)

EdVinyard (approved):
I had only one question, but this looks reasonable to me.

src/components/JsonSchemaViewer.tsx
daniel-white: it did not have the parameter, i added it and didn't set a limit for backwards compatability

PR #258, opened 2024-06-14
chore(deps): upgrade dependencies

daniel-white:
also switch from jest to vitest for less configuration!


matthewmurphy (approved):
Tests look much nicer with vitest and testing-library!

mallachari (approved):
Looks good and still works fine

